PR #9, opened 2018-07-04
OPENEUROPA-582: Update images on docker-compose and drone file.

voidtek:
### Description

Update images on docker-compose and drone file.

### Change log

- Changed: Images on docker-compose and drone file.

### Commands

```sh
no commands

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,11 +47,6 @@
     },
     "extra": {
         "composer-exit-on-patch-failure": true,
-        "patches": {
-            "drupal/core": {
-                "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-04-15/2943172-kernel-test-base-2.patch"
-            }
-        },
         "installer-paths": {
             "build/core": ["type:drupal-core"],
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -48,7 +49,7 @@
         "composer-exit-on-patch-failure": true,
         "patches": {
             "drupal/core": {
-                "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-04-15/2943172-kernel-test-base-2.patch"
+                "https://www.drupal.org/project/drupal/issues/2943172": "https://www.drupal.org/files/issues/2018-07-05/2943172-kernel-test-base-3.patch"
```
pfrenssen: This is changing the patches, this might affect other components that are depending on this. If composer-patches encounters two different versions for the same patch in the dependency tree then it will throw an error.

I would suggest that we wait with this until we were able to remove the patches.

imanoleguskiza: Since we are going to deal with the patches in a follow up I'm approving this PR.

PR #11, opened 2018-08-09
OPENEUROPA-758: Use single entry point object instead of requiring service file

imanoleguskiza:
### Description

Replace old pcas services with single factory usage

### Change log

- Changed: Replace old pcas services with single factory usage


oe_auth.services.yml
```diff
@@ -1,7 +1,9 @@
+parameters:
+  pcas.base_url: ~
```
ademarco: We should prefix parameters and service names with the module's name, something like:
```
parameters:
  oe_auth.pcas.base_url: ~
```

oe_auth.services.yml
```diff
@@ -1,7 +1,9 @@
...
 services:
```
ademarco: Same comment as below for all services exposed by the module.

services.yml.dist
```diff
@@ -1,49 +1,2 @@
...
-
-  pcas.userfactory:
-    class: OpenEuropa\pcas\Security\Core\User\PCasUserFactory
-
-  pcas.serializerfactory:
-    class: OpenEuropa\pcas\Utils\PCasSerializerFactory
-
-  pcas.httpclientfactory:
-    class: OpenEuropa\pcas\Http\HttpClientFactory
+  pcas.base_url: ${authentication.server.base_url}
```
ademarco: Rename this too as per comments above.

PR #15, opened 2018-08-22
OPENEUROPA-1146: Move pcas settings into drupal configuration

imanoleguskiza:
### Description

After the work done in OPENEUROPA-758 the CAS service base URL is provided as a Drupal service parameter. This means that, in order for developers to use the module, they will need to override such service parameter in a custom service.yml file and have that loaded by the $container_yamls array in Drupal's settings.php.

The scope of this ticket is to change that and store the CAS service base URL as Drupal configuration instead.

### Change log

- Added:
- Changed: Moved pcas settings into drupal configuration
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


config/install/oe_authentication.settings.yml
```diff
@@ -0,0 +1 @@
+base_url: http://authentication:8001
```
upchuk: Let's also change the readme to say `authentication` instead of `oeauth-server` for the `etc/hosts` file.

Also, please add to the README that sites can override this base url in their settings.php file.

src/Authentication/Provider/OeAuthentication.php
```diff
@@ -17,23 +17,23 @@ class OeAuthentication implements AuthenticationProviderInterface {
...
-   * @var Drupal\oe_authentication\UserProvider
+   * @var \Drupal\oe_authentication\UserProvider
    */
   protected $userProvider;
 
   /**
    * OeAuthentication constructor.
    *
-   * @param OpenEuropa\pcas\PCasFactory $pCasFactory
+   * @param \OpenEuropa\pcas\PCasFactory $pCasFactory
```
upchuk: This is the `Drupal\oe_authentication\PCasFactory` not the original one.

PR #17, opened 2018-08-23
OPENEUROPA-164: Authentication: fix login error on ECAS redirect

upchuk:
### Description

The problem here is that the CAS response from ECAS does not contain the user email in the assumed `cas:email` key of the XML. The `demo-server` of the PCas library does. 

This PR fixes the error by looking for the user based on the user name found in the `cas:user` key which is part of a standard [CAS response](https://apereo.github.io/cas/5.1.x/protocol/CAS-Protocol-Specification.html#252-response) for ticket validation.

In a follow up we need to ensure that retrieving the email is done for both ECAS and the demo-server.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/src/UserProvider.php
+++ b/src/UserProvider.php
@@ -82,11 +82,12 @@ class UserProvider {
    * @throws \Exception
    */
   protected function doLoadAccount(PCasUserInterface $pCasUser) {
-    $mail = $pCasUser->get('cas:email');
-    if (empty($mail)) {
-      throw new \Exception('Email address not provided by OE Authentication.');
+    $username = $pCasUser->get('cas:user');
+    if (!$username) {
+      throw new \Exception('No username found on the PCas user.');
     }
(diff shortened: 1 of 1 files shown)


src/UserProvider.php
```diff
@@ -82,11 +82,12 @@ protected function attachRoles(UserInterface $account, PCasUserInterface $pCasUs
    * @throws \Exception
    */
   protected function doLoadAccount(PCasUserInterface $pCasUser) {
-    $mail = $pCasUser->get('cas:email');
-    if (empty($mail)) {
-      throw new \Exception('Email address not provided by OE Authentication.');
+    $username = $pCasUser->get('cas:user');
+    if (!$username) {
```
drupol: `if (null === $username)`

drupol:
I would squash this PR, there is no added value to have 2 commits here.

PR #18, opened 2018-08-23
OPENEUROPA-1177: Consolidate oe_authentication

imanoleguskiza:
### Description

Consolidate the oe_authentication Drupal module and its integration with ECAS.

UPDATE: As agreed, we will merge what we had and continue on two other tracks while what we have so far stays in 0.x:

* 1.x - Using the Drupal CAS module
* 2.x - Abstracting the authentication layer to use other authentication systems as well.

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/oe_authentication.module
+++ b/oe_authentication.module
@@ -2,7 +2,7 @@
 
 /**
  * @file
- * OpenEuropa Webtools Laco Widget module.
+ * OpenEuropa Authentication module.
  */
 
 declare(strict_types = 1);
--- a/oe_authentication.module
+++ b/oe_authentication.module
@@ -13,8 +13,8 @@ use Drupal\Core\Form\FormStateInterface;
  * Implements hook_form_alter().
  */
 function oe_authentication_form_user_form_alter(&$form, FormStateInterface $form_state, $form_id) {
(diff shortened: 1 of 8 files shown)


oe_authentication.module
```diff
@@ -0,0 +1,20 @@
+<?php
+
+/**
+ * @file
+ * OpenEuropa Webtools Laco Widget module.
```
upchuk: This is not right :)

tests/Kernel/PCasFactoryTest.php
```diff
@@ -0,0 +1,75 @@
...
+      'allowed_parameters' => ['service', 'renew', 'gateway'],
+    ];
+    $this->assertEquals($login_protocol, $properties['protocol']['login']);
+  }
+
+  /**
+   * Test custom configuration options for the PCasFactory class.
+   */
+  public function testCustomConfiguration(): void {
+    $this->config("oe_authentication.settings")
```
upchuk: Single quotes here please.

PR #20, opened 2018-09-05
OPENEUROPA-784: Make sure component dependencies are as relaxed as possible.

aritomelo:
Components dependencies:

  - Update some version to use tilde, so it uses the minimal minor version and also for a consistency use decimal number as the version.

  - Update some packages to use minimal minor version.

  - Added composer boundary to test lowest and highest.

 - Added packagist badge.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -15,13 +15,30 @@ services:
     image: openeuropa/pcas-server:latest
 
 pipeline:
-  composer-install:
+  composer-install-lowest:
     group: prepare
     image: fpfis/httpd-php-dev:7.1
     volumes:
       - /cache:/cache
     commands:
-      - composer install
+    # @todo remove "composer install" step once the following issue is fixed.
+    # @link https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-1234
+    - composer install --ansi --no-suggest --no-progress
+    - composer update --prefer-lowest --prefer-stable --ansi --no-suggest --no-progress
+    when:
+      matrix:
+        COMPOSER_BOUNDARY: lowest
+
+  composer-install-highest:
+    group: prepare
+    image: fpfis/httpd-php-dev:7.1
+    volumes:
+    - /cache:/cache
+    commands:
+    - composer install --ansi --no-suggest --no-progress
+    when:
+      matrix:
+        COMPOSER_BOUNDARY: highest
 
   site-install:
     image: fpfis/httpd-php-dev:7.1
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -9,17 +9,17 @@
         "drupal/core": "~8.6@alpha",
```
pfrenssen: I would set this to `~8.6` so that it is consistent with the one used for `openeuropa/drupal-core-require-dev`.

.drone.yml
```diff
@@ -15,13 +15,31 @@ services:
...
+  composer-install-lowest:
     group: prepare
     image: fpfis/httpd-php-dev:7.1
     volumes:
       - /cache:/cache
     commands:
-      - composer install
+    # @todo remove "composer install" step once the following issue is fixed.
+    # @link https://webgate.ec.europa.eu/CITnet/jira/browse/OPENEUROPA-1234
+    # @note COMPOSER_ROOT_VERSION , see https://getcomposer.org/doc/articles/troubleshooting.md#package-not-found-on-travis-ci-org
```
voidtek: Why we need this?

composer.json
```diff
@@ -8,20 +8,22 @@
     "require": {
         "php": "^7.1",
         "drupal/cas": "^1.0",
-        "drupal/core": "~8.6@alpha"
+        "drupal/core": "~8.6"
     },
     "require-dev": {
-        "composer/installers": "^1.2",
-        "cweagans/composer-patches": "~1.0",
```
voidtek: We need to keep this dependency.

composer.json
```diff
@@ -8,20 +8,22 @@
...
-        "drupal-composer/drupal-scaffold": "^2.2",
+        "composer/installers": "~1.5",
+        "consolidation/robo": "~1.2.2",
```
voidtek: This dependency is already included in `openeuropa/task-runner`

composer.json
```diff
@@ -8,20 +8,22 @@
...
+        "drupal-composer/drupal-scaffold": "~2.2",
+        "drupal/coder": "~8.2.12",
```
voidtek: This dependency is already included in `openeuropa/drupal-core-require-dev`

composer.json
```diff
@@ -8,20 +8,22 @@
...
-        "nikic/php-parser": "~3",
-        "openeuropa/code-review": "~0.2",
-        "openeuropa/drupal-core-require-dev": "~8.6",
-        "openeuropa/task-runner": "~0.7"
+        "drush/drush": "~9.4",
+        "nikic/php-parser": "~3.0",
+        "openeuropa/code-review": "^1.0.0-alpha4",
+        "openeuropa/drupal-core-require-dev": "^8.6",
+        "openeuropa/task-runner": "~1.0",
+        "php-http/guzzle6-adapter": "~1.0"
```
voidtek: Why we need this dependency?

PR #25, opened 2018-09-27
OPENEUROPA-1206: Removed custom service in favor of new event.

imanoleguskiza:
## OPENEUROPA-1206

In OPENEUROPA-1192, we had to override the CasValidator service in order to apply some EU Login specific data to the validation URL. However, this is not ideal and it would suit other uses cases to make this process alterable in the contributed module.

The scope of this ticket is to create a patch for CAS that improves the getServerValidateUrl() method of the CasValidator (configurable path + dispatches an event to allow modules to alter the params).

### Change log

- Added: New cas PreValidate event subscriber
- Changed:
- Deprecated:
- Removed: Custom EuLoginValidator service
- Fixed:
- Security:


composer.json
```diff
@@ -58,7 +58,7 @@
             "drupal/cas": {
                 "Add support for http CAS servers": "https://www.drupal.org/files/issues/2018-09-03/2996924-4.patch",
                 "Provide custom mail handlers": "https://www.drupal.org/files/issues/2018-09-04/2996955-2.patch",
-                "Provide after validation event": "https://www.drupal.org/files/issues/2018-09-05/2997099-4.patch"
+                "Provide pre validation event": "https://www.drupal.org/files/issues/2018-09-28/3001882-11.patch"
```
upchuk: @imanoleguskiza The patch was merged, so we can remove it from here. 
Same goes for https://www.drupal.org/project/cas/issues/2996924

I think we can use the dev version of CAS, at least until we tag the first 1.0 release. 

@ademarco what do you think? Maybe we can ask him to make a release in the meantime? Last was Feb.

PR #29, opened 2018-10-04
OPENEUROPA-1260: Fix authentication logout redirect for the demo server.

drupol:
### Description

Currently, if logged in with the Pcas Demo server using the oe_authentication component, the logout action fails as no redirect URL can be built to redirect back to (see attachment)

Fix this so that login in with the demo server works. If needed, PR should be opened also on pcas demo.

### DIFF
--- a/oe_authentication.install
+++ b/oe_authentication.install
@@ -18,6 +18,7 @@ function oe_authentication_install() {
   \Drupal::configFactory()->getEditable('cas.settings')
     ->set('server.hostname', 'ecas.ec.europa.eu')
     ->set('server.path', '/cas')
+    ->set('server.version', '3')
     ->set('forced_login.enabled', TRUE)
     ->set('forced_login.paths.pages', '/user/login')
     ->set('logout.cas_logout', TRUE)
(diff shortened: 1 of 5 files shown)


oe_authentication.install
```diff
@@ -18,6 +18,7 @@ function oe_authentication_install() {
   \Drupal::configFactory()->getEditable('cas.settings')
     ->set('server.hostname', 'ecas.ec.europa.eu')
     ->set('server.path', '/cas')
+    ->set('server.version', '3')
```
upchuk: Why are we setting version number 3 here? The default configuration the module installs with is for ECAS. In the runner we override these configuration to make it work on our dev envs with PCAS.

runner.yml.dist
```diff
@@ -35,6 +35,7 @@ commands:
         $config['cas.settings']['server']['hostname'] = 'authentication';
         $config['cas.settings']['server']['port'] = '8001';
         $config['cas.settings']['server']['path'] = '/';
+        $config['cas.settings']['server']['version'] = '3';
```
upchuk: Please see above comment. If version 3 is needed for PCAS, this is the only place it needs to be, not in the .install hook.

oe_authentication.install
drupol: Hello, I tested with both and it works.
For more consistency with the rest, I would leave it like it is.

runner.yml.dist
drupol: As discussed and based on the source of the cas module and pcas server lib, it needs to stay like this to let the module use the proper url argument (url <> service).

PR #30, opened 2018-10-04
OPENEUROPA-1249: Allow access for Drupal users on oe_authentication

imanoleguskiza:
### Description

Allow access for Drupal users on oe_authentication and allow user 1 to create and delete users.

### Change log

- Added: Allow access for Drupal users on oe_authentication and allow user 1 to create and delete users.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:


src/Access/ExternalUserAccessCheck.php
```diff
@@ -0,0 +1,62 @@
...
+   *   The external authentication map.
+   */
+  public function __construct(Authmap $authMap) {
+    $this->authMap = $authMap;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container) {
```
upchuk: What is this? :)

src/Access/ExternalUserAccessCheck.php
```diff
@@ -0,0 +1,62 @@
...
+   */
+  protected $authMap;
+
+  /**
+   * Constructors the ExternalUserAccessCheck.
+   *
+   * @param \Drupal\externalauth\Authmap $authMap
+   *   The external authentication map.
+   */
+  public function __construct(Authmap $authMap) {
```
upchuk: Is there no `AuthMapInterface`?

src/Access/ExternalUserAccessCheck.php
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account) {
```
upchuk: :void

src/Access/SuperUserAccessCheck.php
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account) {
```
upchuk: :void

tests/Behat/OeAuthenticationContext.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_authentication\Behat;
+
+use Drupal\DrupalExtension\Context\ConfigContext;
+
+/**
+ * Defines step definitions specifically for testing the RDF entities.
```
upchuk: Copy paste.

tests/features/drupal-login.feature
```diff
@@ -0,0 +1,11 @@
...
+Feature: Login through Drupal
+  If configured properly
+  I can access the CMS backend through Drupal
+
+  Scenario: If configured properly I can access the CMS backend through Drupal
+    Given the site is configured to use Drupal login
+    And I am logged in as a user with the "authenticated" role
+    Then I should see the link "Log out"
+    When I click "Log out"
+    Then I should not see the link "Log out"
```
upchuk: How sure are we that we are back on the site and everything is fine. Not seeing "Log out" does not mean that we actually logged out right?

src/Access/ExternalUserAccessCheck.php
```diff
@@ -0,0 +1,53 @@
...
+  /**
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account):AccessResultInterface {
```
upchuk: Space after the colon: ':'

src/Access/SuperUserAccessCheck.php
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account):AccessResultInterface {
```
upchuk: Space after the colon: ':'

tests/features/drupal-login.feature
upchuk: Falls outside the scope of this ticket.

src/Routing/RouteSubscriber.php
```diff
@@ -17,21 +17,29 @@ class RouteSubscriber extends RouteSubscriberBase {
    * {@inheritdoc}
    */
   protected function alterRoutes(RouteCollection $collection) {
```
upchuk: Let's put also here :void

src/Access/ExternalUserAccessCheck.php
```diff
@@ -0,0 +1,53 @@
...
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account): AccessResultInterface {
+    $uid = $account->id();
```
drupol: `$uid` is only used once, therefore I would not create a variable for it and refactor the code a bit.
There is also no need to use empty(), we can just use:

```php
    if ([] === $this->authMap->getAll($account->id())) {
```
Regarding the missing empty line before the return, as there is no rule in PHPCS, I wouldn't flag this for now but highligh this during the next retro.

As this is not a blocker nor defined in any PHPCS, it's up to the user to decide if he wants to do it or not, it's not mandatory.

src/Access/SuperUserAccessCheck.php
```diff
@@ -0,0 +1,34 @@
...
+   * Checks access.
+   *
+   * @param \Drupal\Core\Session\AccountInterface $account
+   *   The logged in user key.
+   *
+   * @return \Drupal\Core\Access\AccessResultInterface
+   *   The access result.
+   */
+  public function access(AccountInterface $account): AccessResultInterface {
+    $uid = $account->id();
```
drupol: `$uid` is only used only once, therefore I would not create a variable for it and refactor the code a bit.

```php
    if (1 === $account->id()) {
```

Regarding the missing empty line before the return, as there is no rule in PHPCS, I wouldn't flag this for now but highlight this during the next retro.

As this is not a blocker nor defined in any PHPCS, it's up to the user to decide if he wants to do it or not, it's not mandatory.

src/Routing/RouteSubscriber.php
```diff
@@ -16,22 +16,30 @@ class RouteSubscriber extends RouteSubscriberBase {
...
+
+    // Only the user 1 has access to user creation and deletion on Drupal.
+    $admin_routes = [
       'user.admin_create',
       'user.multiple_cancel_confirm',
+      'user.cancel_confirm',
+    ];
+    foreach ($admin_routes as $admin_route) {
+      if ($route = $collection->get($admin_route)) {
+        $route->setRequirement('_superuser_access_check', 'TRUE');
```
drupol: For better readability, I think it's always better to move away the code from conditions.
It also reduces the cyclomatic complexity in overall.
So, in this case, I think it would be nicer to have:

```php
if (null === ($route = $collection->get($admin_route))) {
  continue;
}
$route->setRequirement('_superuser_access_check', 'TRUE'); 
```

As this is not a blocker nor defined in any PHPCS, it's up to the user to decide if he wants to do it or not, it's not mandatory.

composer.json
```diff
@@ -12,6 +12,7 @@
     },
     "require-dev": {
         "composer/installers": "~1.5",
+        "consolidation/robo": "^1.3",
```
ademarco: We don't need this if we set: 
```
         "openeuropa/task-runner": "~1.0-beta2",
```

behat.yml.dist
```diff
@@ -6,6 +6,7 @@ default:
       contexts:
         - Drupal\DrupalExtension\Context\MinkContext
         - Drupal\DrupalExtension\Context\DrupalContext
+        - Drupal\Tests\oe_authentication\Behat\OeAuthenticationContext
```
ademarco: Convention for Drupal modules is `namespace Drupal\Tests\oe_authentication\...`, I'd also remove the `Oe` in `OeAuthenticationContext` since it's already in the namespace.

PR #34, opened 2018-10-31
OPENEUROPA-1293: Change cas-server to eCAS Mockup Server on Docker Compose.

voidtek:
### Description

Change cas-server to eCAS Mockup Server on Docker Compose.

### Change log

- [x] Configure an eCAS mockup server container with the Official Ecas mockup docker.
- [x] Use this container in oe_authetication instead of the pcas demo server.
- [x] Configure oe_authentication to use this locally via the runner
- [x] Rewrite the behat test


docker-compose.yml.dist
```diff
@@ -18,12 +19,28 @@ services:
...
+      - SE_OPTS=-debug
+    ports:
+      - 5900:5900
+    expose:
+      - 4444
   authentication:
-    image: openeuropa/pcas-server:latest
+    # For local access, run this command in the terminal
+    # echo "127.0.1.1       authentication" >> /etc/hosts
+    image: registry.fpfis.tech.ec.europa.eu/ecas-mock-server:4.6.0
```
imanoleguskiza: My only problem with this is that the container is behind basic authentication so it's not as easy to build a development environment. We should make this clear in the documentation and provide alternatives (like going back the Pol's cas server).

voidtek: The problem here is that the Pol's CAS server doesn't have the basic functionality and it's based in CAS and not in eCAS.
For me, the Pol's CAS Server need to be discontinued.

drupol: hum... what ?
It *has* the very basic functionnality... nothing else for now.

imanoleguskiza: I'm not saying we add it to the repository, but at the very least we should let people know that they won't be able to use the ecas container in their environments and give them examples of other containers that would work (like Pol's)

voidtek: I agree that we need propose some alternative. 
Here, if you want we can propose to use the cas docker container. https://hub.docker.com/r/apereo/cas/
But this container and the Pol's container don't follow the same workflows, rules, endpoints and communication of an eCAS server and according the description of this module ``` The OpenEuropa Authentication module allows to authenticate against EU Login, the European Commission login service.```, we need  to provide a container that simulates the EU Login server.

drupol: Guys, the eCAS is just a EU service implementing the CAS protocol (with very few unnoticeable modifications).
This is why we can use the CAS contrib module for it!

imanoleguskiza: For me it would be enough just to add a comment so that potential users are aware of the limitation :)

imanoleguskiza: @voidtek will take care of this on OPENEUROPA-1299

voidtek:
See https://github.com/openeuropa/oe_authentication/pull/35

PR #35, opened 2018-11-16
OPENEUROPA-1299: Override proxyCallback for ECAS.

voidtek:
### Description

Override proxyCallback for ECAS.

### Change log

- [x] Configure an eCAS mockup server container with the Official Ecas mockup docker.
- [x] Use this container in oe_authetication instead of the pcas demo server.
- [x] Configure oe_authentication to use this locally via the runner
- [x] Rewrite the behat test
- [x] Override the proxyCallback function
- [x] Adding kernel test for the proxyCallback
- [x] Documentation oe_authentication functionnality
- [x] Documentation oe_authentication docker/image


README.md
```diff
@@ -59,8 +119,8 @@ to `./runner.yml` and overriding relevant properties.
 
 To set up the project run:
 
-```
-$ ./vendor/bin/run drupal:site-setup
+```bash
+./vendor/bin/run drupal:site-setup
```
dxvargas: This is already run as post command of composer.

src/Controller/ProxyCallbackController.php
```diff
@@ -0,0 +1,124 @@
...
+    $pgt_iou = $current_request->request->get('pgtIou');
+
+    // Check for both a pgtIou and pgtId parameter.
+    // If either is not present, inform EULogin Server of an error.
+    if (!isset($pgt_id) && !isset($pgt_iou)) {
+      return new Response(t('Missing necessary parameters'), 400);
+    }
+
+    if ($this->checkPgtMapping($pgt_iou, $pgt_id)) {
+      return new Response(t('Parameters already use'), 400);
```
dxvargas: Should be "Parameters already **in** use".

src/Controller/ProxyCallbackController.php
```diff
@@ -0,0 +1,124 @@
...
+   *   Return TRUE is present in the database.
+   */
+  protected function checkPgtMapping($pgt_iou, $pgt_id) {
+    $result = $this->connection->select('cas_pgt_storage')
+      ->fields('cas_pgt_storage', ['pgt_iou', 'pgt'])
+      ->condition('pgt_iou', $pgt_iou)
+      ->condition('pgt', $pgt_id)
+      ->execute()->fetchAll();
+
+    return count($result) > 0 ? TRUE : FALSE;
```
dxvargas: More simple:

`return count($result) > 0;`

tests/Kernel/ProxyCallbackTest.php
```diff
@@ -0,0 +1,78 @@
...
+    $xml = '<?xml version="1.0" encoding="UTF-8"?><proxySuccess xmlns="https://localhost"/>';
+    $this->assertEquals($xml, $response->getContent());
+
+    // Check a duplicate.
+    $test_request = Request::create($proxy_callback_url, 'POST');
+    $test_request->request->set('pgtId', 'foo');
+    $test_request->request->set('pgtIou', 'bar');
+    $response = $kernel->handle($test_request);
+    $this->assertEquals(400, $response->getStatusCode(), $response->getContent());
+    $this->assertEquals('Parameters already use', $response->getContent());
```
dxvargas: Should be "Parameters already in use" (modify only if it was modified in the tested code, of course).

src/Controller/ProxyCallbackController.php
```diff
@@ -0,0 +1,124 @@
...
+
+    $response = new Response($xml);
+    $response->headers->set('Content-Type', 'xml');
+
+    return $response;
+  }
+
+  /**
+   * Check if the pgtIou and pgtId are mapped in the database.
+   *
```
dxvargas: It would be interesting to mention here that if this check is not made, a 500 error is thrown. And also a @todo to put this check as a patch in the parent class in the contrib module.

src/Controller/ProxyCallbackController.php
```diff
@@ -0,0 +1,124 @@
...
+
+  /**
+   * Route callback for the ProxyGrantingTicket information.
+   *
+   * This function stores the incoming PGTIOU and pgtId parameters so that
+   * the incoming response from the EULogin Server can be looked up.
+   */
+  public function callback() {
+    $config = $this->configFactory->get('oe_authentication.settings');
+    if ($config->get('protocol') === 'cas') {
```
imanoleguskiza: This sounds strange. Since we are the ones using a specific protocol, should we not check if $config->get('protocol') !== 'eulogin' ?

README.md
voidtek: The all paragraph makes sens, see:

> Before setting up and installing the site make sure to customize default configuration values by copying [runner.yml.dist](runner.yml.dist)
> to `./runner.yml` and overriding relevant properties.
> 
> To set up the project run:
> 
> ```bash
> ./vendor/bin/run drupal:site-setup
> ```

src/Controller/ProxyCallbackController.php
voidtek: This will be my participation for the contrib module.

src/Controller/ProxyCallbackController.php
voidtek: Make totally sense.

PR #38, opened 2018-12-12
OPENEUROPA-1485: Add drupal patch to prevent blocked users from loggi…

imanoleguskiza:
…ng in.

### Description

Users loging in with ecas get a success message even when they don't log in. A patch was proposed to fix this.

### Change log

- Added: Drupal 3020014 patch added


imanoleguskiza:
Drupal patch was accepted so I'm closing the PR

PR #39, opened 2018-12-13
OPENEUROPA-1486: Remove access to user delete.

imanoleguskiza:
### Description

A user that is deleted on the site will be recreated as soon as he logs in again using EU Login. The scope of this issue is to disable the ability to delete a user, even for user 1.

### Change log

- Removed: Access to user delete

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "drupal/devel": "~1.2",
         "drupal/drupal-extension": "~4.0",
         "drush/drush": "~9.0@stable",
+        "egulias/email-validator": "^1.2.1",
         "nikic/php-parser": "~3.0",
         "openeuropa/behat-transformation-context" : "~0.1",
         "openeuropa/code-review": "~1.0.0-alpha4",
--- a/oe_authentication.module
+++ b/oe_authentication.module
@@ -7,14 +7,12 @@
 
 declare(strict_types = 1);
 
-use Drupal\Core\Form\FormStateInterface;
-
 /**
- * Implements hook_form__FORM_ID_alter().
+ * Implements hook_user_cancel_methods_alter().
  *
  * Remove the option to delete users while keeping the option to block them.
  */
-function oe_authentication_form_user_cancel_form_alter(&$form, FormStateInterface $form_state) {
+function oe_authentication_user_cancel_methods_alter(&$methods) {
   $user = \Drupal::currentUser();
   if (!$user->hasPermission('delete user')) {
     $restricted_options = [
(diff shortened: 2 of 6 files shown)


oe_authentication.module
```diff
@@ -6,3 +6,20 @@
...
+
+use Drupal\Core\Form\FormStateInterface;
+
+/**
+ * Implements hook_form__FORM_ID_alter().
+ *
+ * Remove the option to delete users while keeping the option to block them.
+ */
+function oe_authentication_form_user_cancel_form_alter(&$form, FormStateInterface $form_state) {
+  $restricted_options = [
```
upchuk: Let's introduced a new permission and only apply these restrictions if the user does not have this permission.

tests/features/user-management.feature
```diff
@@ -0,0 +1,21 @@
+@api @javascript
+Feature: Manage users
+  In order to manage the users of the site
+  As an administrator with the appropriate rights
+  I need to be able to edit and delete users
+
+  Scenario: A user with the appropriate rights can delete users.
+    Given the site is configured to use Drupal login
+    And I am logged in as a user with the "administer users, delete user" permissions
```
brummbar: It looked to me that the test doesn't cover enough, so I went and tried do delete a user through the bulk action "Cancel the selected user accounts". 
That form `UserMultipleCancelConfirm` is indeed not covered. I also noticed that the deletion methods are retrieved from a function `user_cancel_methods()`, which gives a centralised place where to alter the available methods.

I also noticed that by having a user with these two permissions above and trying to actually delete the user causes a 403, as a few routes have been disabled for non-superadmin users (while the `entity.user.cancel_form` has been left out). Thus I'm not sure if having the `delete user` permission is useful.

upchuk: I agree. 

1. The cancel methods are properly altered though as the `hook_user_cancel_methods_alter()` is being used. So this is the way to go.
2. Removed permission. We've already made the decision earlier when blocking these routes but I forgot. So it's my bad that i threw this confusion to add a permission.

drupol:
Just rebased against master to fix conflicts.

composer.json
```diff
@@ -18,6 +18,7 @@
         "drupal/devel": "~1.2",
         "drupal/drupal-extension": "~4.0",
         "drush/drush": "~9.0@stable",
+        "egulias/email-validator": "^1.2.1",
```
drupol: I have concerns about this.

First, if adding this line is fixing a bigger issue with user forms, then it's out of the scope of this and should be on its own sub ticket and this ticket should depend on it.

upchuk: Isnt this for the lowest dependency thing? @drupol @imanoleguskiza

drupol: Yes it is. It fixes an issue with user forms when built using lowest deps. I think it should be tackled in a ticket on its own.

PR #40, opened 2018-12-17
OPENEUROPA-1485: Prevent blocked users from logging in through cas

imanoleguskiza:
### Description

Add tests to make sure users logging throuhg ecas can't log in

### Change log

- Added: Tests for blocked users

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,6 +53,11 @@
             "build/profiles/contrib/{$name}": ["type:drupal-profile"],
             "build/modules/contrib/{$name}": ["type:drupal-module"],
             "build/themes/contrib/{$name}": ["type:drupal-theme"]
+        },
+        "patches": {
+            "drupal/cas": {
+                "https://www.drupal.org/project/cas/issues/3020014": "https://www.drupal.org/files/issues/2018-12-12/3020014-2.patch"
+            }
         }
     },
     "config": {
--- a/tests/Behat/DrupalContext.php
+++ b/tests/Behat/DrupalContext.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_authentication\Behat;
+
+use Drupal\DrupalExtension\Context\DrupalContext as BaseDrupalContext;
+
+/**
+ * Defines step definitions specifically for testing the CAS users.
+ */
+class DrupalContext extends BaseDrupalContext {
+
+  /**
+   * Configures the CAS module to use Drupal login.
+   *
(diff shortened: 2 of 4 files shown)


composer.json
```diff
@@ -53,6 +53,11 @@
             "build/profiles/contrib/{$name}": ["type:drupal-profile"],
             "build/modules/contrib/{$name}": ["type:drupal-module"],
             "build/themes/contrib/{$name}": ["type:drupal-theme"]
+        },
+        "patches": {
+            "drupal/cas": {
```
upchuk: Why you adding the patch? We are using dev version of the module and the patch was merged no?

tests/Behat/DrupalContext.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_authentication\Behat;
+
+use Drupal\DrupalExtension\Context\DrupalContext as BaseDrupalContext;
+
+/**
+ * Defines step definitions specifically for testing the CAS users.
+ */
+class DrupalContext extends BaseDrupalContext {
```
upchuk: We already have an AuthenticationContext, why create a new one? Also, we should never extend from another Context. Only from RawContexts..

tests/Behat/DrupalContext.php
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Configures the CAS module to use Drupal login.
```
upchuk: This entire method docblock and definition is copy pasted.

tests/Behat/DrupalContext.php
```diff
@@ -0,0 +1,34 @@
...
+   *   The name of the user to be blocked.
+   *
+   * @When the user :username is blocked
+   *
+   * @throws \Exception
+   *   Thrown when the user with the given name does not exist.
+   */
+  public function setConfigDrupalLogin($username): void {
+    /** @var \Drupal\user\Entity\User $user */
+    $user = user_load_by_name($username);
```
upchuk: Why are you using global functions?

tests/features/ecas-login.feature
```diff
@@ -34,3 +34,30 @@ Feature: Login through OE Authentication
     And I should not see the link "My account"
     And I should not see the link "Log out"
     And I should see the link "Log in"
+
+  Scenario: A blocked user should not be able to log in
+    When I am on the homepage
```
upchuk: Is the first part of the scenario needed? We are testing the successful login in the previous scenario no?

PR #45, opened 2019-01-29
OPENEUROPA-1520: Use drupal core.

imanoleguskiza:
### Description

Remove drupal-core-require and use drupal/core
### Change log

- Added:
- Changed: Remove drupal-core-require and use drupal/core
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "prefer-stable": true,
     "require": {
         "drupal/cas": "dev-1.x",
-        "openeuropa/drupal-core-require": "^8.6",
+        "drupal/core": "^8.6",
         "php": "^7.1"
     },
     "require-dev": {


vever001:
Hello everyone,

Ensuing https://github.com/openeuropa/drupal-core-require/issues/9
This should solve the issue I'm having but I have another question:
I noticed all openeuropa modules use "drupal/core": "^8.6" as a requirement. Which is there only to build drupal and run the tests no?
Is there another possibility so that the project owners are still fully in control on the drupal core version and could even have the freedom to use an old drupal core version if they wanted to (say 8.5.x)?

Thanks,
Hervé

brummbar:
Hello @vever001, the drupal/core version in the requirements ensures that our component is compatible with the versions covered by the constraints. Project owners won't have the possibility to lower that requirement, like they cannot do for other libraries (unless using clever trickeries).
They can use older releases of our components and handle the gaps with the missing functionality by themselves, if they really need an older version of core.
Why would be needed to stay on an older core version? Considering the fact that minor releases don't cause backwards compatibility breaks?

PR #47, opened 2019-02-11
OPENEUROPA-1586: Add extra EULogin fields.

imanoleguskiza:
### Description

Provide extra fields for the user entity to hold the following information coming from EU Login:
- First name
- Last name
- Department
- Organisation

### Change log

- Added: Extra EULogin fields.


tests/features/ecas-login.feature
```diff
@@ -25,6 +25,13 @@ Feature: Login through OE Authentication
     When I click "My account"
     Then I should see the heading "chucknorris"
 
+    # Profile contains extra fields.
+    When I click "Edit"
+    Then the "First Name" field should contain "Chuck"
+    Then the "Last Name" field should contain "NORRIS"
```
nagyad: This and the rest should be "And"

oe_authentication.install
```diff
@@ -38,4 +41,64 @@ function oe_authentication_install() {
 
     \Drupal::messenger()->addMessage('Please be aware that oe_authentication makes changes to the following configuration: CAS settings.');
   }
+
+  _oe_update_eulogin_fields();
+}
+
+/**
+ * Implements hook_update_N().
```
nagyad: I think we don't need this. At least core doesn't do it. Also in drush it will print it, so I think it is better to leave informative comment here only.

src/Event/EuLoginEventSubscriber.php
```diff
@@ -68,16 +68,22 @@ public static function getSubscribedEvents(): array {
...
-    if (!empty($attributes['mail'])) {
-      $event->setPropertyValue('mail', $attributes['mail']);
+    if (!empty($attributes['email'])) {
+      $event->setPropertyValue('mail', $attributes['email']);
     }
-
-    if (!empty($attributes['authenticationFactors'])) {
-      $authFactors = $attributes['authenticationFactors'];
-      if (isset($authFactors['moniker'])) {
-        $event->setPropertyValue('mail', $authFactors['moniker']);
```
upchuk: Are you sure this is correct? Before, the email was coming inside `moniker` attributes. How come this changed? Is it the `userDetails` that makes this change?

imanoleguskiza: The problem with this snippet is that its not accurate: moniker is the string used to log in, be it an email or a username. This lead to wrong information being stored in the database. With the userDetails enabled, the email always comes with the "email" property. I could add an exception if no email is found to simply prevent a user from being created, would this make sense?

oe_authentication.module
```diff
@@ -21,3 +24,47 @@ function oe_authentication_user_cancel_methods_alter(&$methods) {
...
+ *
+ * Add custom EULogin fields.
+ */
+function oe_authentication_entity_base_field_info(EntityTypeInterface $entity_type) {
+  if ($entity_type->id() != 'user') {
+    return [];
+  }
+  $custom_fields = [
+    '7' => [
+      'machine_name' => 'field_oe_firstname',
```
upchuk: Shall we add also some descriptions to these fields?

PR #54, opened 2019-02-28
OPENEUROPA-1685: Add post login event.

imanoleguskiza:
### Description

User information from EULogin needs to be updated on every login
### Change log

- Added: User update on login.
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


src/Event/EuLoginEventSubscriber.php
```diff
@@ -54,36 +55,69 @@ public static function create(ContainerInterface $container) {
...
+   *   An array of attributes retrieved from CAS.
+   *
+   * @return array
+   *   An array of field names and values.
+   */
+  private function generateUserDataArray(array $attributes) {
+    $properties = [];
     if (!empty($attributes['email'])) {
-      $event->setPropertyValue('mail', $attributes['email']);
+      $properties['mail'] = $attributes['email'];
```
pfrenssen: This is doing a conversion of CAS properties to fields on the user object. This seems like something that can be useful to be reused by other code. Isn't there a better place for this than locked away as an uncallable private method on an event subscriber?

tests/features/ecas-login.feature
```diff
@@ -42,16 +42,47 @@ Feature: Login through OE Authentication
...
-      | name      |
-      | lsalander |
-    And the user "lsalander" is blocked
-    When I am on the homepage
-    Then I should see the link "Log in"
-    And I should not see the link "Log out"
+    Given the user "lsalander" is blocked
 
     # When I try to log in again I will be denied access.
+    When I am on the homepage
```
pfrenssen: A precondition for the test starts with the keyword `@Given` and not `@When`.

How to know when it is a precondition or a vital part of the test - you can ask yourself if this step is essential for the scenario being tested. I.e.

Is it essential for a blocked user to be on the homepage in order to not be able to log in? No -> `@Given`, Yes -> `@When`.

tests/features/ecas-login.feature
```diff
@@ -42,16 +42,47 @@ Feature: Login through OE Authentication
...
+
   Scenario: A blocked user should not be able to log in
-    Given users:
-      | name      |
-      | lsalander |
-    And the user "lsalander" is blocked
-    When I am on the homepage
-    Then I should see the link "Log in"
-    And I should not see the link "Log out"
+    Given the user "lsalander" is blocked
```
pfrenssen: This might not work any more, the user `lsalander` doesn't exist at this point in the scenario so can't be blocked.

It is usually a good idea not to reuse user names etc from previous tests. Since these features can become very long with a large number of scenarios it can be difficult to figure out which exact test fails if they all use the same test data.

src/CasProcessor.php
```diff
@@ -0,0 +1,97 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_authentication;
+
+/**
+ * Helper functions to process Cas attributes.
+ */
+class CasProcessor {
+
+  const USER_CAS_ATTRIBUTE_MAPPING = [
```
pfrenssen: Please document this constant.

src/CasProcessor.php
```diff
@@ -0,0 +1,97 @@
...
+   *   An array containing the parsed attributes.
+   */
+  public static function processValidationResponseAttributes(string $source): array {
+    $dom = new \DOMDocument();
+    $dom->preserveWhiteSpace = FALSE;
+    $dom->encoding = "utf-8";
+
+    // Suppress errors from this function, as we intend to allow other
+    // event subscribers to work on the data.
+    if (@$dom->loadXML($source) === FALSE) {
```
pfrenssen: I'm not 100% sure on this section, in the context of the event subscriber this made sense, but in this class we can reasonably expect to be passed in a valid processable string.

I would suggest to provide an additional static method `isValidationResponse(string $response): bool` that contains this check and the `authenticationSuccess` one. Then this can be called from the event subscriber.

This would also make it clean, having the validation and processing in separate, reusable methods :heart_eyes:

drupol: And on top of that, suppressing any potential error with '@' is not a practice that we want to use here I guess...

pfrenssen: I agree that suppressing errors is a bad practice, but the DOM extension is a very special case and here it is fine I think. The DOM extension is not a "native" PHP extension but it is wrapping the `libxml` library which is very chatty about errors when parsing, and the extension just bubbles up anything reported by `libxml`, even if the document ends up to be parsed adequately. It would just cause noise in our logs and possibly on the user's screen.

Suppressing the errors but checking the output and throwing an exception if anything goes wrong would also allow us to track any problems, but in a way that is not needlessly chatty about XML specifics :)

tests/features/ecas-login.feature
imanoleguskiza: The problem is that we can't use any user since they have to come from the EULogin instance. Currently the only two users available are the ones we are already using. I think finding a solution for this issue is out of the scope for this ticket but I acknoledge the problem you presented so I added extra steps to make sure the user is available.

pfrenssen: I cannot access this EULogin image, but can't we create test users in it? Having to rely on hardcoded user names seems pretty limiting, especially for testing purposes. We definitely need to have the ability to run tests against dozens of users, and users with special characters in their names etc. Anyway this is out of scope.

src/CasProcessor.php
```diff
@@ -0,0 +1,122 @@
...
+    // Suppress errors from this function, as we intend to allow other
+    // event subscribers to work on the data.
+    if (@$dom->loadXML($response) === FALSE) {
+      return FALSE;
+    }
+
+    $success_elements = $dom->getElementsByTagName("authenticationSuccess");
+    if ($success_elements->length === 0) {
+      return FALSE;
+    }
```
pfrenssen: This can be simplified to `return $success_elements->length !== 0;`.

src/CasProcessor.php
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Check whether the validation response is valid or not.
+   *
+   * @param string $response
+   *   The response to be validated.
+   *
+   * @return bool
+   *   Whether the validation response is valid or not.
+   */
+  public static function isValidationResponse(string $response) {
```
pfrenssen: I would rename this to `isValidResponse()`, that is clearer.

tests/features/ecas-login.feature
```diff
@@ -42,16 +42,59 @@ Feature: Login through OE Authentication
...
+
   Scenario: A blocked user should not be able to log in
-    Given users:
-      | name      |
-      | lsalander |
-    And the user "lsalander" is blocked
-    When I am on the homepage
-    Then I should see the link "Log in"
-    And I should not see the link "Log out"
+    # Login with an EULogin user.
```
pfrenssen: Explain "why" these steps are taken, instead of "what" is being done. For example:

```
# Check that a user which is not blocked can log in normally.
```

pfrenssen:
I reviewed the code and it looks pretty good. I just have some few small remarks. I wasn't able to test it though, I just did a code review.

imanoleguskiza:
Addressed the comments :)

PR #56, opened 2019-03-04
OPENEUROPA-1649: eulogin default behavior on first login

imanoleguskiza:
### Description

Block users if administrator approval is required. Allow users to be created in Drupal if appropriate permissions are set.

### Change log

- Added:
- Changed: Block users if administrator approval is required.
- Deprecated:
- Removed: Restriction on user creation
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```


src/Event/EuLoginEventSubscriber.php
```diff
@@ -19,6 +21,7 @@
  * the required modifications to work with EU Login.
  */
 class EuLoginEventSubscriber implements EventSubscriberInterface {
+  use StringTranslationTrait;
```
upchuk: Empty line needed.

src/Event/EuLoginEventSubscriber.php
```diff
@@ -68,6 +81,14 @@ public static function getSubscribedEvents(): array {
    */
   public function generateEmail(CasPreRegisterEvent $event): void {
     $attributes = $event->getCasPropertyBag()->getAttributes();
+    $user_settings = $this->configFactory->get('user.settings');
```
upchuk: This method has gone way past its respos. So let's either rename it to be more generic or break it up into multiple ones.

tests/features/ecas-login.feature
```diff
@@ -43,17 +44,13 @@ Feature: Login through OE Authentication
     And I should see the link "Log in"
 
   Scenario: A blocked user should not be able to log in
```
upchuk: The scenario name no longer reflects what we test. Now we test that by default, any new users are created as blocked.

tests/features/ecas-login.feature
```diff
@@ -43,17 +44,13 @@ Feature: Login through OE Authentication
...
 
     # When I try to log in again I will be denied access.
-    When I click "Log in"
+    When I am on the homepage
+    And I click "Log in"
     And I fill in "Username or e-mail address" with "Lisbeth.SALANDER@ext.ec.europa.eu"
     And I fill in "Password" with "dragon_tattoo"
     And I press the "Login!" button
     Then I should see "There was a problem logging in, please contact a site administrator."
+    And I should see "Thank you for applying for an account. Your account is currently pending approval by the site administrator."
```
upchuk: We are not actually asserting the user is blocked, only that s/he sees the message. Let's add a check the ensure the user is actually blocked.

dxvargas: Added a step "And I should see the link "Log in"" that ensures user is actually blocked, marking as resolved.

tests/features/ecas-login.feature
dxvargas: Updated to "A site that requires administration validation on users should block them by default", marking the conversation as resolved.

src/Event/EuLoginEventSubscriber.php
dxvargas: Method was renamed to "processUserProperties", conversation can be resolved.

src/Event/EuLoginEventSubscriber.php
dxvargas: Done.

PR #61, opened 2019-04-04
Issue 60: Prevent the usage of http-foundation latest version.

imanoleguskiza:
### Description

After the latest symfony/http-foundation 3.4.24 update Drupal core has reported an issue with its lazy sessions: https://www.drupal.org/project/drupal/issues/3045349

The current solution is to fix the http-foundation version in te composer.json file:

"conflict": { "symfony/http-foundation": "3.4.24" },


hernani:
Not needed anymore.

PR #67, opened 2019-05-02
OPENEUROPA-1713: Adjust/correct behat tests of component.

sergepavle:
### Description

Currently we are able to have green behat tests only at first time. On other we are getting failed scenarios. It is related to fact that we are creating/blocking users and didn't remove/unblock after scenario.
+ Aligning with ticket OPENEUROPA-1840

### Change log

- Added:
- Changed:
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/tests/Behat/AuthenticationContext.php
+++ b/tests/Behat/AuthenticationContext.php
@@ -138,7 +138,7 @@ class AuthenticationContext extends RawDrupalContext {
   }
 
   /**
-   * We reset the authentication mock to the state it was when we found it.
+   * We reset the authentication mock to the state as it was.
    *
    * @AfterScenario @ecas-login
    */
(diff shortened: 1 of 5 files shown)


tests/features/ecas-login.feature
```diff
@@ -40,11 +41,49 @@ Feature: Login through OE Authentication
...
+    And I fill in "Username or e-mail address" with "007@mi6.eu"
+    And I fill in "Password" with "shaken_not_stirred"
+    And I press the "Login!" button
+    # Redirected back to Drupal.
+    Then I should see "You have been logged in."
+    And I should see the link "My account"
+    And I should see the link "Log out"
+    And I should not see the link "Log in"
+    # Redirected back to Drupal.
+
```
nagyad: Empty line after comment.

tests/Behat/AuthenticationContext.php
```diff
@@ -137,4 +137,41 @@ public function setNewUsersBlocked(): void {
     $this->configContext->setConfig('user.settings', 'register', USER_REGISTER_VISITORS_ADMINISTRATIVE_APPROVAL);
   }
 
+  /**
+   * We reset the authentication mock to the state it was when we found it.
```
nagyad: We reset the authentication mock to the state as it was.

tests/Behat/AuthenticationContext.php
```diff
@@ -137,4 +137,41 @@ public function setNewUsersBlocked(): void {
...
+
+  /**
+   * Allow external users to login.
+   *
+   * @BeforeScenario @AllowExternalLogin
+   */
+  public function allowExternalUsers(): void {
+    // Set the assurance level to allow also external users to login. Security
+    // implications are mitigated by the fact that users are by default blocked
+    // when they login for the first time if they don't yet exist.
```
upchuk: This part is not needed in the comment:

```
Security
    // implications are mitigated by the fact that users are by default blocked
    // when they login for the first time if they don't yet exist.
```

tests/Behat/AuthenticationContext.php
```diff
@@ -137,4 +137,41 @@ public function setNewUsersBlocked(): void {
...
+    $config = \Drupal::configFactory()->getEditable('oe_authentication.settings');
+    $config->set('assurance_level', 'LOW');
+    $config->save();
```
upchuk: Can't we used the gathered ConfigContext and set this like that? This way we don't have to revert it ourselves in an after scenario?

PR #85, opened 2019-06-04
Maje /user/register redirect optional

### LINKED ISSUE #83: The /user/register route alway redirects to EU Login
This should be configurable as some sites may want to offer also Drupal registration. Other sites might decide to show a page with several authentication providers/authorities.

claudiu-cristea:
Fixes #83

### DIFF
--- a/behat.yml.dist
+++ b/behat.yml.dist
@@ -16,8 +16,9 @@ default:
               user registration: '/user/register'
   extensions:
     Drupal\MinkExtension:
-      goutte: ~
-      selenium2: ~
+      goutte:
+        guzzle_parameters:
+          verify: false
       ajax_timeout: 10
       browser_name: 'chrome'
       javascript_session: 'selenium2'
--- a/config/install/oe_authentication.settings.yml
+++ b/config/install/oe_authentication.settings.yml
@@ -3,3 +3,4 @@ register_path: 'eim/external/register.cgi'
 validation_path: 'TicketValidationService'
 assurance_level: TOP
 ticket_types: SERVICE,PROXY
+redirect_user_register_route: true
--- a/config/schema/oe_authentication.schema.yml
+++ b/config/schema/oe_authentication.schema.yml
@@ -17,3 +17,6 @@ oe_authentication.settings:
     ticket_types:
       type: string
       label: 'Application available ticket types'
+    redirect_user_register_route:
+      type: boolean
+      label: 'Redirect Drupal user registration route to EU Login'
--- a/src/Routing/RouteSubscriber.php
+++ b/src/Routing/RouteSubscriber.php
@@ -4,6 +4,7 @@ declare(strict_types = 1);
 
 namespace Drupal\oe_authentication\Routing;
 
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Routing\RouteSubscriberBase;
 use Symfony\Component\Routing\Route;
 use Symfony\Component\Routing\RouteCollection;
@@ -12,6 +13,23 @@ use Symfony\Component\Routing\RouteCollection;
  * Listens to route events.
  */
 class RouteSubscriber extends RouteSubscriberBase {
+
+  /**
+   * The config factory service.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInterface
+   */
+  protected $configFactory;
+
+  /**
+   * Constructs a new route event subscriber.
+   *
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   The config factory service.
+   */
+  public function __construct(ConfigFactoryInterface $config_factory) {
+    $this->configFactory = $config_factory;
+  }
 
   /**
    * {@inheritdoc}
@@ -33,12 +51,15 @@ class RouteSubscriber extends RouteSubscriberBase {
     }
 
     // Replace the core register route controller.
-    $route = $collection->get('user.register');
-    if ($route instanceof Route) {
-      $defaults = $route->getDefaults();
-      unset($defaults['_form']);
-      $defaults['_controller'] = '\Drupal\oe_authentication\Controller\RegisterController::register';
-      $route->setDefaults($defaults);
+    $config = $this->configFactory->get('oe_authentication.settings');
+    if ($config->get('redirect_user_register_route')) {
+      $route = $collection->get('user.register');
+      if ($route instanceof Route) {
+        $defaults = $route->getDefaults();
+        unset($defaults['_form']);
+        $defaults['_controller'] = '\Drupal\oe_authentication\Controller\RegisterController::register';
+        $route->setDefaults($defaults);
+      }
     }
 
     // Replace the cas callback route controller.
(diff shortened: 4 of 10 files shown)


src/Routing/RouteSubscriber.php
```diff
@@ -33,12 +51,15 @@ protected function alterRoutes(RouteCollection $collection): void {
...
-    $route = $collection->get('user.register');
-    if ($route instanceof Route) {
-      $defaults = $route->getDefaults();
-      unset($defaults['_form']);
-      $defaults['_controller'] = '\Drupal\oe_authentication\Controller\RegisterController::register';
-      $route->setDefaults($defaults);
+    $config = $this->configFactory->get('oe_authentication.settings');
+    if ($config->get('redirect_user_register_route')) {
+      $route = $collection->get('user.register');
+      if ($route instanceof Route) {
```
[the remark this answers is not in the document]
pfrenssen: I made the issue: #86

tests/Behat/AuthenticationContext.php
```diff
@@ -105,6 +104,29 @@ public function backupCasConfigs(): void {
...
+   * the routes at the end of the scenario for tests that requesting it, by
+   * using the @RebuildRouter tag.
+   *
+   * @see \Drupal\oe_authentication\Event\UserRegisterRouteRedirectConfigSubscriber
+   * @see \Drupal\DrupalExtension\Context\ConfigContext::cleanConfig()
+   *
+   * @AfterScenario @RebuildRouter
+   */
+  public function rebuildRoutes(): void {
+    \Drupal::service('router.builder')->rebuild();
```
pfrenssen: I am not sure that this is the right approach. If we are changing a config value in a test and this will persist in the following tests then we are violating the principle that each test should be standalone and not affect any other tests.

In reality, any tests that depend on the default value will start failing once we change this value. Meaning that these tests might run fine the first time, but will fail if we run them a second time.

I think we should remove this section and instead revert the configuration change at the end of the test.

tests/features/ecas-register.feature
```diff
@@ -1,12 +1,22 @@
-@javascript @ecas-login
+@api @ecas-login
 Feature: Register through OE Authentication
   In order to be able to have new users
   As an anonymous user of the system
   I need to be able to go to the registration URL
 
+  @DrupalLogin @BackupAuthConfigs @RebuildRouter
```
pfrenssen: OK I see here that there is this `@BackupAuthConfigs` tag which presumably takes care of restoring the configuration.

Can we maybe extend this hook to rebuild the router?

I don't really like this approach of changing configuration through the UI, and then relying on tags to magically restore the values, but if this is the existing convention followed by the project then it is better to follow it.

But let's try not to end up with more tags than necessary :)

tests/features/ecas-register.feature
```diff
@@ -1,12 +1,22 @@
...
     Then I should see "Create an account"
+
+    Given I am logged in as a user with the "administer authentication configuration" permission
+    When I am on "the Authentication configuration page"
+    And I uncheck "Redirect user registration route to EU Login"
+    Then I press "Save configuration"
+
+    Given I am an anonymous user
+    When I visit "the user registration page"
+    Then I should see "Create new account"
```
pfrenssen: See this is what I mean. We are changing the value from ON to OFF in the UI, but now we do not have a test that checks if it also works correctly if we change it from OFF to ON. This is a UI test so it makes sense we reopen the form to check that the updated value is displayed correctly and can be toggled back. It can happen that there is a bug with the "default_value" in the form.

This is why I think that using tags to "fix" changes made in the UI are pointless. A complete UI test can restore the values through the UI.

behat.yml.dist
```diff
@@ -16,8 +16,9 @@ default:
               user registration: '/user/register'
   extensions:
     Drupal\MinkExtension:
-      goutte: ~
-      selenium2: ~
+      goutte:
+        guzzle_parameters:
+          verify: false
```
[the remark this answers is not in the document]
claudiu-cristea: I made the test not to use Selenium (removed the `@javascript` tag). The test makes no JavaScript testing and should not be a JS test. But now, as is using Guzzle to make requests, that one is throwing `GuzzleHttp\Exception\RequestException` exceptions because the requests are HTTPS but the eCAS Mock Sever uses a self-signed certificate. See https://stackoverflow.com/questions/35637140/behat-selenium-tests-trust-self-signed-ssl-cerificate

config/install/oe_authentication.settings.yml
```diff
@@ -3,3 +3,4 @@ register_path: 'eim/external/register.cgi'
 validation_path: 'TicketValidationService'
 assurance_level: TOP
 ticket_types: SERVICE,PROXY
+redirect_user_register_route: true
```
idimopoulos: Do we need an update path for this? In the route subscriber, it does nothing if the setting is not set and this is set to true to maintain the default functionality.

config/schema/oe_authentication.schema.yml
```diff
@@ -16,4 +16,7 @@ oe_authentication.settings:
       label: 'Application assurance levels'
     ticket_types:
       type: string
-      label: 'Application available ticket types'
\ No newline at end of file
+      label: 'Application available ticket types'
+    redirect_user_register_route:
```
idimopoulos: The `cas` module already offers a similar setting called `prevent_normal_login`. Maybe for consistency we could have the same as `prevent_normal_register`.
Also, that would be a good contribution to cas itself instead of forcing the redirection in the `oe_authentication` module but it seems that they are not very eager to do so since there are responses like this https://www.drupal.org/project/cas/issues/2884216

config/install/oe_authentication.settings.yml
claudiu-cristea: It's just config management. When the config import will find a diff between installed `oe_authentication.settings` and the code version, will apply the code version, right?

config/schema/oe_authentication.schema.yml
claudiu-cristea: This is "case close" according to https://www.drupal.org/project/cas/issues/2884216. The redirect is provided by `oe_authentication`, not `cas`. So, we have to fix the problem here, not in `cas`.

config/install/oe_authentication.settings.yml
idimopoulos: That would assume that a site or a distribution imports all settings with every update and also handle config sync conflicts. Normally you would expect that a new version of a module that has new settings will take care on its own to perform the necessary changes in the database.

claudiu-cristea: Ok, added an update path.

behat.yml.dist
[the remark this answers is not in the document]
pfrenssen: Oh right! That makes sense, thanks!

HermanScis:
@hernani I think I need the code in here but this pull request has been approved a year ago and is still sitting here... Any plans with this module?

HermanScis:
I've tried to revolve the merge conflict in my fork https://github.com/HermanScis/oe_authentication/tree/83-registrer-redirect See  #122

PR #87, opened 2019-06-05
Allow user admins to cancel accounts with deletion

### LINKED ISSUE #79: Why only superuser is able to cancel accounts with deletion?
See `oe_authentication_user_cancel_methods_alter()`. Why only superuser is able to cancel accounts with deletion?

Looks like this restriction in hard coupled with some OE business logic (?)

claudiu-cristea:
Fixes #79


oe_authentication.module
```diff
@@ -16,7 +16,7 @@ use Drupal\Core\Field\BaseFieldDefinition;
  * Remove the option to delete users while keeping the option to block them.
  */
 function oe_authentication_user_cancel_methods_alter(&$methods) {
-  if (\Drupal::currentUser()->id() == 1) {
+  if (\Drupal::currentUser()->hasPermission('administer users')) {
```
brummbar: We cannot use this permission as we assign it to some roles but we don't want them to be able to do those type of actions. Let's create a permission "delete users" and use that.

claudiu-cristea: @brummbar yeh, I was askin' this in #79

brummbar: My bad, missed the notification :(

claudiu-cristea:
After discussions with @brummbar and analysing the whole issue, it's clear to me that blocking the deletion makes no sense:

- On many sites `user_cancel_delete` is the default site method and users cannot select the cancellation method. As an effect, a regular user will get a horrible error message.
- An idea came up from @brummbar to allow deletion for users granted with `cancel account` permission. Making the check `$account->hasPermission('cancel account')` will allow deletion only for UID1 and users with `cancel account` permission. But, anyway, if a user doesn't have the `cancel account` permission he cannot cancel at all, so cannot access the "Cancel account" button. As a conclusion, it makes no sense to check this condition.

From a functional perspective, it doesn't seem a good idea to block the ability of a user to delete his account. Looking from a GDPR angle, it's not legal to keep user data against user's will. But if you still have good reasons to do so, than, fine, **revoke its `cancel account` permissions**.

What I'm proposing in the last commit is removing this limitation. The logic is too coupled to some OpenEuropa functionality and doesn't look as a feature of a reusable module. IMHO, it should not live here, but in the project that uses this functionality. Joinup is the proof of this fact.

But if you decide not to merge this removal, we'll be forced to workaround by implementing `hook_module_implements_alter()` and removing the `oe_authentication_user_cancel_methods_alter` from the list of implementations. So ugly!

claudiu-cristea:
This is a "won't fix".

PR #89, opened 2019-06-14
Move user fields into a sub-module

### LINKED ISSUE #81: Don't create and map user fields. Use cas_attributes instead
When installed, the module creates next base fields for the `user` entity: `field_oe_firstname`, `field_oe_lastname`, `field_oe_department`, `field_oe_organisation`. Also, on registering a new Drupal account with data taken from EU Login, it fills those fields with data.

This approach is hard coupled to a specific OE business model, preventing module reusability. The module should not be in business of configuring the `user` entity fields. That should be deferred to other module, to the parent project or to configuration management (if the fields are converted from base to configurable).

Also there's no way to allow users change data on the site, e.g family name, because everytime they are logging in, data from EU Login is overwriting the local changes.

### Proposal

* Stop adding `user` base fields.
* Use `cas_attributes` module to map and set fields.

claudiu-cristea:
Fixes #81

# QA notes

## Clean install

Is covered by the Behat tests.

## Testing the update path

Install de development environment according to the README.md.

```bash
git checkout master
./vendor/bin/run drupal:site-setup
./vendor/bin/run drupal:site-install
./vendor/bin/drush -l http://localhost/path/to/site/ uli
```

Go to `/admin/reports/status` and check that there are no Entity/Field Definitions mismatch errors.

Do a login for a test user (see the eCAS Mock Server users data) and, in the `users_field_data` table, check that the profile fields (`field_oe_*`) are correctly filled.

```bash
git checkout 81-user-fields
./vendor/bin/drush updb -y
```

Refresh `/admin/reports/status` and check that there are no Entity/Field Definitions mismatch errors.

Check the `users_field_data` table. The fields should be filled with the previous data.

### DIFF
--- a/modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
+++ b/modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
@@ -10,7 +10,7 @@ namespace Drupal\oe_authentication_user_fields;
 class EuLoginAttributesHelper {
 
   /**
-   * Array mapping EU Login attributes with user account fields.
+   * Array mapping of EU Login attributes with user account fields.
    */
   const USER_EU_LOGIN_ATTRIBUTE_MAPPING = [
     'mail' => 'email',
@@ -21,13 +21,13 @@ class EuLoginAttributesHelper {
   ];
 
   /**
-   * Converts an array EU Login attributes into an array of Drupal field/values.
+   * Converts the EU Login attributes into a Drupal field/values array.
    *
    * @param array $attributes
    *   An array containing a series of EU Login attributes.
    *
    * @return array
-   *   An array containing a series of Drupal field names and values.
+   *   An associative array of field values indexed by the field name.
    */
   public static function convertEuLoginAttributesToFieldValues(array $attributes): array {
     $values = [];
--- a/modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
+++ b/modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_authentication_user_fields;
+
+/**
+ * Helper functions to map EU Login attributes to user profile.
+ */
+class EuLoginAttributesHelper {
+
+  /**
+   * Array mapping EU Login attributes with user account fields.
+   */
+  const USER_EU_LOGIN_ATTRIBUTE_MAPPING = [
+    'mail' => 'email',
+    'field_oe_firstname' => 'firstName',
+    'field_oe_lastname' => 'lastName',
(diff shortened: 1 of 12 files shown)


modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
```diff
@@ -0,0 +1,42 @@
...
+  const USER_EU_LOGIN_ATTRIBUTE_MAPPING = [
+    'mail' => 'email',
+    'field_oe_firstname' => 'firstName',
+    'field_oe_lastname' => 'lastName',
+    'field_oe_department' => 'departmentNumber',
+    'field_oe_organisation' => 'domain',
+  ];
+
+  /**
+   * Converts an array EU Login attributes into an array of Drupal field/values.
```
idimopoulos: `Converts an array of* EU Login....`

modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
```diff
@@ -0,0 +1,42 @@
...
+   *
+   * @param array $attributes
+   *   An array containing a series of EU Login attributes.
+   *
+   * @return array
+   *   An array containing a series of Drupal field names and values.
```
idimopoulos: Isn't it more usual in Drupal to describe this as something like "An associative array of field values indexed by the field name"?

modules/oe_authentication_user_fields/src/EventSubscriber/EuLoginAttributesToUserFieldsSubscriber.php
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  /**
+   * Updates the user data based on the information taken from EU Login.
+   *
+   * @param \Drupal\cas\Event\CasPostLoginEvent $event
+   *   The triggered event.
+   *
+   * @throws \Drupal\Core\Entity\EntityStorageException
+   *   In case of failures an exception is thrown.
```
idimopoulos: The exception class seems to specific for a broad exception message.

idimopoulos: I am removing this entirely referring back to a discussion between Antonio and Pieter about not trying to document or catch exceptions related to the entity storage operations unless we are passing them as parameters when retrieving them.

modules/oe_authentication_user_fields/src/EuLoginAttributesHelper.php
idimopoulos: I changed this a bit since it was exceeding the 80 lines limit.

idimopoulos:
I discussed with @claudiu-cristea and since he is already occupied with another task and these are simply nitpicks, I would implement them after him agreeing with them.

PR #92, opened 2019-06-18
Add the /eulogin route as a separate instance that doesn't replace the default route

### LINKED ISSUE #90: Don't alter routes from the CAS module
For some reason the module is altering the path of the `cas.login` route to `/eulogin`. This is breaking the functionality as documented in the CAS module. There is no value in doing this.

The code responsible for this is as follows, as you can see there is no reason given to explain why this breaking change is required:

```
  protected function alterRoutes(RouteCollection $collection): void {
    // ...

    // Replace the cas callback route controller.
    if ($route = $collection->get('cas.proxyCallback')) {
      $route->setDefaults([
        '_controller' => '\Drupal\oe_authentication\Controller\ProxyCallbackController::callback',
      ]);
    }

    // Replace default cas login route with eulogin one.
    if ($route = $collection->get('cas.login')) {
      $route->setPath('/eulogin');
    }
  }
```

Also no reason is given for altering the `proxyCallback` controller.

Please make sure that these paths are made available without breaking the default URLs provided by the CAS module.

pfrenssen:
Fixes #90

### DIFF
--- a/oe_authentication.routing.yml
+++ b/oe_authentication.routing.yml
@@ -1,3 +1,11 @@
+oe_authentication.login:
+  path: '/eulogin'
+  defaults:
+    _controller: '\Drupal\cas\Controller\ForceLoginController::forceLogin'
+  requirements:
+    _access: 'TRUE'
+  options:
+    _maintenance_access: TRUE
 oe_authentication.settings:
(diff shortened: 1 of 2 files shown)


pfrenssen:
It is failing due to a missing MySQL instance on the test bot.

PR #99, opened 2019-11-15
OPENEUROPA-2502: Add param groups to validation path.

dxvargas:
### Description

Add param groups to validation path to get user groups.

### Change log

- Added: Parameter groups=* in `\Drupal\oe_authentication\Event\EuLoginEventSubscriber::alterValidationPath`


src/CasProcessor.php
```diff
@@ -87,7 +87,12 @@ private static function parseAttributes(\DOMElement $node): array {
       else {
         $value = $child->nodeValue;
       }
-      $attributes[$name] = $value;
+      if ($name === 'group') {
```
upchuk: Can't we figure out if the attribute is an array without hardcoding the name of it?  Also, isn't the `group` attribute one array level down?

Also, no tests for this?

src/CasProcessor.php
```diff
@@ -72,31 +72,60 @@ public static function processValidationResponseAttributes(string $source): arra
...
+   */
+  protected static function isAssociative(\DOMElement $node): bool {
+    $names = [];
+    foreach ($node->childNodes as $key => $child) {
+      $names[] = $child->localName;
+    }
+
+    // We consider it as associative if we have only one name value or if  the
+    // name values are different.
+    return count($names) < 2 || count(array_unique($names)) > 1;
```
imanoleguskiza: Not sure if this is possible but if we have some children with the same name and some others with different ones then they will be processed as associative and the children with the same name will be overwriten.

src/CasProcessor.php
```diff
@@ -72,31 +72,60 @@ public static function processValidationResponseAttributes(string $source): arra
...
+   * @return bool
+   *   Whether the node children should be mapped to an associated array.
+   */
+  protected static function isAssociative(\DOMElement $node): bool {
+    $names = [];
+    foreach ($node->childNodes as $key => $child) {
+      $names[] = $child->localName;
+    }
+
+    // We consider it as associative if we have only one name value or if  the
```
imanoleguskiza: Double space before the "the"

src/CasProcessor.php
```diff
@@ -72,31 +72,60 @@ public static function processValidationResponseAttributes(string $source): arra
...
+   *
+   * @param \DOMElement $node
+   *   The node element.
+   *
+   * @return bool
+   *   Whether the node children should be mapped to an associated array.
+   */
+  protected static function isAssociative(\DOMElement $node): bool {
+    $names = [];
+    foreach ($node->childNodes as $key => $child) {
```
imanoleguskiza: No need to define the $key variable

src/CasProcessor.php
dxvargas: Fixed.

PR #131, opened 2021-06-09
EWPP-1062: Handle group property in user data.

sergepavle:


brummbar (changes requested):
I was happy about the PR, but then I noticed the double quotes...
image

brummbar (approved):

brummbar (approved):

imanoleguskiza (approved):

PR #133, opened 2021-06-30
OEL-153: Prevent CAS from creating a new user account if the email is already taken.

escuriola:
### Description

Not allow to register a user from CAS if a user with the same email exists locally. Inform of this error with a friendly message.

### Change log

- Added: 
- Changed: Modifications to an already subscribed events
- Deprecated:
- Removed:
- Fixed:
- Security:

### Commands

```sh
[Insert commands here]

```

### DIFF
--- a/src/Event/EuLoginEventSubscriber.php
+++ b/src/Event/EuLoginEventSubscriber.php
@@ -86,6 +86,23 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
     if ($user_settings->get('register') === UserInterface::REGISTER_VISITORS_ADMINISTRATIVE_APPROVAL) {
       $event->setPropertyValue('status', 0);
     }
+    // If user from cas does not exists and user_accounts.auto_register is set,
+    // then check if user mail exists before create.
+    // If exists, user can not be registered neither login
+    // and must be informed with a message.
+    $cas_settings = $this->configFactory->get('cas.settings');
+    $auto_register = $cas_settings->get('user_accounts.auto_register');
+    if ($auto_register === TRUE) {
+      $bag = $event->getCasPropertyBag();
+      $email = $bag->getAttribute('email');
+
+      $userMailAlreadyExists = user_load_by_mail($email);
+      if ($userMailAlreadyExists !== FALSE) {
+        $event->cancelAutomaticRegistration('A user with this mail already exists. Please contact with your site administrator.');
+        $event->setPropertyValue('status', 'error_automatic_registration');
+      }
+    }
+
   }
 
   /**
--- a/src/Event/EuLoginEventSubscriber.php
+++ b/src/Event/EuLoginEventSubscriber.php
@@ -67,10 +67,31 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
    */
   public static function getSubscribedEvents(): array {
     $events = [];
-    $events[CasHelper::EVENT_PRE_REGISTER] = 'processUserProperties';
+    $events[CasHelper::EVENT_PRE_REGISTER] = [
+      ['checkUserMailExists', 1000],
+      ['processUserProperties'],
+    ];
     $events[CasHelper::EVENT_POST_VALIDATE] = 'processCasAttributes';
     $events[CasHelper::EVENT_PRE_VALIDATE] = 'alterValidationPath';
     return $events;
+  }
+
+  /**
+   * Checks user e-mail exists previously.
+   *
+   * @param \Drupal\cas\Event\CasPreRegisterEvent $event
+   *   The triggered event.
+   */
+  public function checkUserMailExists(CasPreRegisterEvent $event): void {
+    $cas_settings = $this->configFactory->get('cas.settings');
+    if ($cas_settings->get('user_accounts.auto_register')) {
+      $email = $event->getCasPropertyBag()->getAttribute('email');
+
+      if (user_load_by_mail($email) !== FALSE) {
+        $event->cancelAutomaticRegistration(t('A user with this mail already exists. Please contact with your site administrator.')->render());
+        $event->stopPropagation();
+      }
+    }
   }
 
   /**
@@ -86,23 +107,6 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
     if ($user_settings->get('register') === UserInterface::REGISTER_VISITORS_ADMINISTRATIVE_APPROVAL) {
       $event->setPropertyValue('status', 0);
     }
-    // If user from cas does not exists and user_accounts.auto_register is set,
-    // then check if user mail exists before create.
-    // If exists, user can not be registered neither login
-    // and must be informed with a message.
-    $cas_settings = $this->configFactory->get('cas.settings');
-    $auto_register = $cas_settings->get('user_accounts.auto_register');
-    if ($auto_register === TRUE) {
-      $bag = $event->getCasPropertyBag();
-      $email = $bag->getAttribute('email');
-
-      $userMailAlreadyExists = user_load_by_mail($email);
-      if ($userMailAlreadyExists !== FALSE) {
-        $event->cancelAutomaticRegistration('A user with this mail already exists. Please contact with your site administrator.');
-        $event->setPropertyValue('status', 'error_automatic_registration');
-      }
-    }
-
   }
 
   /**
--- a/src/Event/EuLoginEventSubscriber.php
+++ b/src/Event/EuLoginEventSubscriber.php
@@ -77,7 +77,7 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
   }
 
   /**
-   * Checks user e-mail exists previously.
+   * Checks user email exists previously.
    *
    * @param \Drupal\cas\Event\CasPreRegisterEvent $event
    *   The triggered event.
@@ -87,8 +87,8 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
     if ($cas_settings->get('user_accounts.auto_register')) {
       $email = $event->getCasPropertyBag()->getAttribute('email');
 
-      if (user_load_by_mail($email) !== FALSE) {
-        $event->cancelAutomaticRegistration(t('A user with this mail already exists. Please contact with your site administrator.')->render());
+      if (user_load_by_mail($email)) {
+        $event->cancelAutomaticRegistration($this->t('A user with this mail already exists. Please contact with your site administrator.'));
         $event->stopPropagation();
       }
     }
--- a/src/Event/EuLoginEventSubscriber.php
+++ b/src/Event/EuLoginEventSubscriber.php
@@ -88,7 +88,7 @@ class EuLoginEventSubscriber implements EventSubscriberInterface {
       $email = $event->getCasPropertyBag()->getAttribute('email');
 
       if (user_load_by_mail($email)) {
-        $event->cancelAutomaticRegistration($this->t('A user with this mail already exists. Please contact with your site administrator.'));
+        $event->cancelAutomaticRegistration($this->t('A user with this email address already exists. Please contact the site administrator.'));
         $event->stopPropagation();
       }
     }
--- a/src/Event/MessengerEuLoginEventSubscriber.php
+++ b/src/Event/MessengerEuLoginEventSubscriber.php
@@ -62,6 +62,10 @@ class MessengerEuLoginEventSubscriber implements EventSubscriberInterface {
       return;
     }
 
+    if ($properties['status'] === 'error_automatic_registration') {
+      return;
+    }
+
     if (!$properties['status']) {
       $this->messenger->addStatus($this->t('Thank you for applying for an account. Your account is currently pending approval by the site administrator.'));
     }
--- a/src/Event/MessengerEuLoginEventSubscriber.php
+++ b/src/Event/MessengerEuLoginEventSubscriber.php
@@ -62,10 +62,6 @@ class MessengerEuLoginEventSubscriber implements EventSubscriberInterface {
       return;
     }
 
-    if ($properties['status'] === 'error_automatic_registration') {
-      return;
-    }
-
     if (!$properties['status']) {
       $this->messenger->addStatus($this->t('Thank you for applying for an account. Your account is currently pending approval by the site administrator.'));
     }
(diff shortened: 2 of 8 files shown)


escuriola:
Behat tests will fail until https://www.drupal.org/project/cas/issues/3221111

claudiu-cristea:
> Behat tests will fail until [drupal.org/project/cas/issues/3221111](https://www.drupal.org/project/cas/issues/3221111)

You should apply the patch from https://www.drupal.org/project/cas/issues/3221111 in composer.json of this module.

src/Event/EuLoginEventSubscriber.php
```diff
@@ -86,6 +86,23 @@ public function processUserProperties(CasPreRegisterEvent $event): void {
...
+    // and must be informed with a message.
+    $cas_settings = $this->configFactory->get('cas.settings');
+    $auto_register = $cas_settings->get('user_accounts.auto_register');
+    if ($auto_register === TRUE) {
+      $bag = $event->getCasPropertyBag();
+      $email = $bag->getAttribute('email');
+
+      $userMailAlreadyExists = user_load_by_mail($email);
+      if ($userMailAlreadyExists !== FALSE) {
+        $event->cancelAutomaticRegistration('A user with this mail already exists. Please contact with your site administrator.');
```
claudiu-cristea, line 101: Well... the reason is the message shown to an end-user, so we should translate it.

src/Event/MessengerEuLoginEventSubscriber.php
```diff
@@ -62,6 +62,10 @@ public function showUserMessage(CasPreRegisterEvent $event): void {
       return;
     }
 
+    if ($properties['status'] === 'error_automatic_registration') {
+      return;
+    }
```
claudiu-cristea, line 67: See above.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
...
 
+  Background:
+    Given CAS users:
+      | Username    | E-mail                            | Password           | First name | Last name | Department    | Organisation |
+      | jb007       | 007@mi6.eu                        | shaken_not_stirred | James      | Bond      | DIGIT.A.3.001 | external     |
+
   Scenario: Register
     Given I am an anonymous user
-    When I visit "the user registration page"
+    When I visit "cas-mock-server/eim/external/register.cgi"
```
claudiu-cristea, line 14: I don't see the reason of this change.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
...
     # Redirected to the Ecas mockup server.
     Then I should see "Create an account"
     And I should see "This page is part of the EU login mock."
+
+  @cleanup:user
+  Scenario: Register new user with AutoRegister enabled
+    Given the site is configured to register users if not exists
```
claudiu-cristea, line 21: Behat tests are mean to end-users. Instead of doing this, we should login as a user with CAS settings permissions, do the settings using UI, then logout.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
...
+    And I fill in "Username or e-mail address" with "007@mi6.eu"
+    And I fill in "Password" with "shaken_not_stirred"
+    And I press the "Login!" button
+    # Redirected back to Drupal.
+    Then I should see "Your account is currently pending approval by the site administrator."
+    And I should see the link "Log in"
+
+  @cleanup:user
+  Scenario: Register with AutoRegister enabled a user with already existent e-mail
+    Given the site is configured to register users if not exists
```
claudiu-cristea, line 34: Same.

tests/Behat/AuthenticationContext.php
```diff
@@ -147,4 +147,30 @@ public function allowExternalUsers(): void {
     $this->configContext->setConfig('oe_authentication.settings', 'assurance_level', 'LOW');
   }
 
+  /**
+   * Configures the CAS server so users are created in login if not exists.
+   *
+   * @Given the site is configured to register users if not exists
```
claudiu-cristea, line 153: We don't need this step. We should do all operations via UI. This the reason why Behat testing exists.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
...
+    Given a user with the same email already exists locally
```
claudiu-cristea, line 35: Why? We should only move the Background user creation here and add before `Create users:`. See my comment from `AuthenticationContext::createLocalUser()`.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
+@api @casMockServer
 Feature: Register through OE Authentication
   In order to be able to have new users
   As an anonymous user of the system
   I need to be able to go to the registration URL
 
+  Background:
+    Given CAS users:
+      | Username    | E-mail                            | Password           | First name | Last name | Department    | Organisation |
+      | jb007       | 007@mi6.eu                        | shaken_not_stirred | James      | Bond      | DIGIT.A.3.001 | external     |
```
claudiu-cristea, line 10: This create the user for all tests. It shouldn't be like this. Move it to the test that really needs it.

tests/Behat/AuthenticationContext.php
```diff
@@ -147,4 +147,30 @@ public function allowExternalUsers(): void {
...
+   */
+  public function theSiteIsConfiguredToRegisterUsersIfNotExists(): void {
+    $this->configContext->setConfig('user_accounts.auto_register', 'cas.settings', TRUE);
+  }
+
+  /**
+   * Create a local user.
+   *
+   * @Given a user with the same email already exists locally
```
claudiu-cristea, line 162: We don't need this. Just use "Given users:", create a local use with this email. Then use "Given CAS users" (see my comment in the .feature file) and add a `Local username` column to link the local user. For details, see `CasMockServerContext::registerUsers()`.

tests/features/ecas-register.feature
```diff
@@ -1,11 +1,44 @@
...
+    Given the site is configured to register users if not exists
+    Given a user with the same email already exists locally
+    When I am on the homepage
+    And I click "Log in"
+    # Redirected to the mock server.
+    And I fill in "Username or e-mail address" with "007@mi6.eu"
+    And I fill in "Password" with "shaken_not_stirred"
+    And I press the "Login!" button
+    # Redirected back to Drupal.
+    Then I should see "A user with this mail already exists. Please contact with your site administrator."
```
claudiu-cristea, line 43: We can do better: `Then I should see the error message "..."`

claudiu-cristea (changes requested):
Here's a first review. Apart from the code remarks, I wonder if we need a setting in this module, something like: _Forbid auto-registration of users with the same email_, defaulting to TRUE. But please ask the team that requested this fix if they need such a switch. It's a business logic decision.

tests/features/ecas-register.feature
escuriola: In version 1.6 this test failed locally

src/Event/EuLoginEventSubscriber.php
```diff
@@ -67,12 +67,33 @@ public static function create(ContainerInterface $container) {
...
+   *
+   * @param \Drupal\cas\Event\CasPreRegisterEvent $event
+   *   The triggered event.
+   */
+  public function checkUserMailExists(CasPreRegisterEvent $event): void {
+    $cas_settings = $this->configFactory->get('cas.settings');
+    if ($cas_settings->get('user_accounts.auto_register')) {
+      $email = $event->getCasPropertyBag()->getAttribute('email');
+
+      if (user_load_by_mail($email) !== FALSE) {
```
claudiu-cristea, line 90: Don't need the `!== FALSE`, if a value is tested as a boolean, PHP will cast it to a boolean. In this case, if `user_load_by_mail()` returns a UserInterface object it will cast it to `TRUE`.

src/Event/EuLoginEventSubscriber.php
```diff
@@ -67,12 +67,33 @@ public static function create(ContainerInterface $container) {
...
+        $event->cancelAutomaticRegistration(t('A user with this mail already exists. Please contact with your site administrator.')->render());
```
claudiu-cristea, line 91: Using the procedural `t()` is discouraged. You can use `$this->t()`, as `StringTranslationTrait` trait is already used.

tests/features/ecas-register.feature
```diff
@@ -9,3 +10,40 @@ Feature: Register through OE Authentication
     # Redirected to the Ecas mockup server.
     Then I should see "Create an account"
     And I should see "This page is part of the EU login mock."
+
+  @cleanup:user
+  Scenario: Login user with auto-register enabled
+    Given I am an anonymous user
+    Given CAS users:
```
claudiu-cristea, line 17: Normally, if we have many `Given`s, only the 1st one is `Given`, the following should be `And`. Also, I would swap the order because `I am an anonymous user` starts the UI journey while `CAS users` is using API to create the CAS users.

tests/features/ecas-register.feature
```diff
@@ -9,3 +10,40 @@ Feature: Register through OE Authentication
...
+    And I should see the link "Log in"
+
+  @cleanup:user
+  Scenario: Login user with an already registered email with auto-register enabled
+    Given I am an anonymous user
+    Given users:
+      | name    | mail        |
+      | james   | 007@mi6.eu  |
+
+    Given CAS users:
```
claudiu-cristea, line 37: Same: Give > And > And. Same observation `Given I am an anonymous user` should be the last.

tests/features/ecas-register.feature
```diff
@@ -9,3 +10,40 @@ Feature: Register through OE Authentication
...
+      | Username    | E-mail                            | Password           | First name | Last name | Department    | Organisation |
+      | jb007       | 007@mi6.eu                        | shaken_not_stirred | James      | Bond      | DIGIT.A.3.001 | external     |
+    When I am on the homepage
+    And I click "Log in"
+    # Redirected to the mock server.
+    And I fill in "Username or e-mail address" with "007@mi6.eu"
+    And I fill in "Password" with "shaken_not_stirred"
+    And I press the "Login!" button
+    # Redirected back to Drupal.
+    Then I should see "Your account is currently pending approval by the site administrator."
```
claudiu-cristea, line 27: Maybe we can use here a more exact assertion: _Then I should see the success message "..."_ (still, I don't remember if is success or warning - to be tested).

src/Event/EuLoginEventSubscriber.php
```diff
@@ -67,12 +67,33 @@ public static function create(ContainerInterface $container) {
...
+      ['checkUserMailExists', 1000],
+      ['processUserProperties'],
+    ];
     $events[CasHelper::EVENT_POST_VALIDATE] = 'processCasAttributes';
     $events[CasHelper::EVENT_PRE_VALIDATE] = 'alterValidationPath';
     return $events;
   }
 
+  /**
+   * Checks user e-mail exists previously.
```
claudiu-cristea, line 80: s/'Checks'/'Checks that'. Also there is a Drupal convention to use 'email' rather than 'e-mail'. See https://www.drupal.org/drupalorg/style-guide/content#relatedwords

claudiu-cristea (changes requested):
I like how this is shaping. Functionally, everything seems to be in place, I only requesting some cosmetic changes.

src/Event/EuLoginEventSubscriber.php
```diff
@@ -67,12 +67,33 @@ public static function create(ContainerInterface $container) {
...
+   * @param \Drupal\cas\Event\CasPreRegisterEvent $event
+   *   The triggered event.
+   */
+  public function checkUserMailExists(CasPreRegisterEvent $event): void {
+    $cas_settings = $this->configFactory->get('cas.settings');
+    if ($cas_settings->get('user_accounts.auto_register')) {
+      $email = $event->getCasPropertyBag()->getAttribute('email');
+
+      if (user_load_by_mail($email)) {
+        $event->cancelAutomaticRegistration($this->t('A user with this mail already exists. Please contact with your site administrator.')->render());
```
claudiu-cristea, line 91: I didn't saw the `->render()` part. That should be removed. We can pass a MarkupInterface to the method, don't need to be string (unless there's something wrong in other place).

claudiu-cristea (dismissed):
Approving. Thank you!

src/Event/EuLoginEventSubscriber.php
```diff
@@ -67,12 +67,33 @@ public static function create(ContainerInterface $container) {
...
+   * @param \Drupal\cas\Event\CasPreRegisterEvent $event
+   *   The triggered event.
+   */
+  public function checkUserMailExists(CasPreRegisterEvent $event): void {
+    $cas_settings = $this->configFactory->get('cas.settings');
+    if ($cas_settings->get('user_accounts.auto_register')) {
+      $email = $event->getCasPropertyBag()->getAttribute('email');
+
+      if (user_load_by_mail($email)) {
+        $event->cancelAutomaticRegistration($this->t('A user with this mail already exists. Please contact with your site administrator.'));
```
bramtenhove, line 91:
```suggestion
        $event->cancelAutomaticRegistration($this->t('A user with this email address already exists. Please contact your site administrator.'));
```

tests/features/ecas-register.feature
```diff
@@ -9,3 +10,38 @@ Feature: Register through OE Authentication
...
+      | jb007       | 007@mi6.eu     | shaken_not_stirred | James      | Bond      | DIGIT.A.3.001 | external     |
+    And I am an anonymous user
+    When I am on the homepage
+    And I click "Log in"
+    # Redirected to the mock server.
+    And I fill in "Username or e-mail address" with "007@mi6.eu"
+    And I fill in "Password" with "shaken_not_stirred"
+    And I press the "Login!" button
+    # Redirected back to Drupal.
+    Then I should see the error message "A user with this mail already exists. Please contact with your site administrator."
```
bramtenhove, line 46:
```suggestion
    Then I should see the error message "A user with this email address already exists. Please contact your site administrator."
```

src/Event/EuLoginEventSubscriber.php
bramtenhove: A suggestion for a small improvement for the text.

claudiu-cristea (approved):
Approved.

brummbar (changes requested):

claudiu-cristea (approved):
Looks OK

brummbar (approved):

brummbar (approved):

claudiu-cristea:
@alesrebec use CAS 2.x

PR #153, opened 2022-05-02
OE-AUTH-01: Sanitize users one by one instead of loading them in bulk…

FGDumitru:
… (memory consumption issue).

### Description

On sites with a big number of users (100K+), the user Drush sanitization function from the oe_authentication_user_fields module is trying to load the entire list of users in memory before sanitizating them and that, in turn, depending on the PHP CLI ram limits, might trigger an Out of memory error.

Issue triggered by this: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FPFISSUPP-16923

### Change log

- Added:
- **Changed**: Changed the users loading to a one by one basis in order to not generate an Out of memory error.
- Deprecated:
- Removed:
- Fixed:
- Security:


FGDumitru:
Superseeded by https://github.com/openeuropa/oe_authentication/pull/154

PR #154, opened 2022-05-02
FPFISSUPP-16923: Use directly a SQL query for the sanitization.

voidtek:
### Description

Use directly a SQL query for the sanitization.

### Change log

- Changed: Sanitization of the users.


gervasek (approved):

sergepavle (changes requested):
Tests should be updated as this change will lead to failure in tests.
Also, it is better to keep field value unique.

voidtek:
> Tests should be updated as this change will lead to failure in tests. Also, it is better to keep the field value unique (https://www.drupal.org/docs/drupal-apis/database-api/update-queries). Probably we my use ->expression() method.

Feel free to copy my patch and apply best practices.

voidtek:
MR followed by https://github.com/openeuropa/oe_authentication/pull/155

PR #171, opened 2023-02-08
EWPP-2984: Drupal 10 compatibility.

brummbar:
Also lower dependencies tests are dropped, as the maintenance of dependencies derived from other components became too complex. It will be restored when we can target only our defined dependencies.


drishu (approved):

.drone.yml
```diff
@@ -93,9 +76,10 @@ pipeline:
...
 
 matrix:
-  COMPOSER_BOUNDARY:
-    - lowest
-    - highest
-  PHP_VERSION:
-    - 8.0
-    - 8.1
+  include:
+    - CORE_VERSION: 9.4
```
drishu, line 80: and then here we go for 9.4.0 and the `~` will install the latest stable or we could also do 9.4.* I think

drishu (changes requested):

drishu (approved):

composer.json
```diff
@@ -90,7 +70,8 @@
             "cweagans/composer-patches": true,
             "dealerdirect/phpcodesniffer-composer-installer": true,
             "drupal/core-composer-scaffold": true,
-            "phpro/grumphp": true
+            "phpro/grumphp": true,
+            "phpstan/extension-installer": true
```
sergepavle, line 74:
```suggestion
            "phpro/grumphp": true
```
Are we really `phpstan/extension-installer` use that?

sergepavle: I did executions:
`docker-compose exec web composer require drupal/core:~9.4.0 drupal/core-composer-scaffold:~9.4.0 --update-with-all-dependencies --ansi --no-progress --no-interaction`
`docker-compose exec web composer require drupal/core:~10.0 drupal/core-composer-scaffold:~10.0 --update-with-all-dependencies --ansi --no-progress --no-interaction`
with PHP8.1/8.0 and didn't notice that we need that.

brummbar: It's coming from core-dev 10.x, I added it in a previous test where I attempted to add back core-dev. I'll remove it.

sergepavle (approved):

drishu (approved):

PR #203, opened 2025-02-07
OPES-1546: Allow to configure user conditions for 2FA

### LINKED ISSUE #202: Allow to configure user conditions for 2FA
In the oe_authentication settings form:

1. Retrieve all the condition plugins that require a user entity.
2. Expose their subform in the above form (like the block form does).
3. Save the data inside the configuration. Check to save the data only if fields are filled in.
4. Add tests.

It's not in scope of this PR to apply the plugin behavior.

brummbar:
Fixes #202 .

### DIFF
--- a/src/Form/AuthenticationSettingsForm.php
+++ b/src/Form/AuthenticationSettingsForm.php
@@ -106,6 +106,7 @@ class AuthenticationSettingsForm extends ConfigFormBase {
     $form['condition_tabs'] = [
       '#type' => 'vertical_tabs',
       '#title' => $this->t('Two-factor authentication conditions'),
+      '#description' => $this->t('Two-factor authentication will be required to log in <strong>only if at least one condition</strong> successfully matches the account that is attempting to log in. Conditions apply only if two-factor authentication is enabled.'),
       '#parents' => ['condition_tabs'],
       '#states' => [
         'visible' => [
--- a/src/Form/AuthenticationSettingsForm.php
+++ b/src/Form/AuthenticationSettingsForm.php
@@ -107,6 +107,11 @@ class AuthenticationSettingsForm extends ConfigFormBase {
       '#type' => 'vertical_tabs',
       '#title' => $this->t('Two-factor authentication conditions'),
       '#parents' => ['condition_tabs'],
+      '#states' => [
+        'visible' => [
+          ':input[name="force_2fa"]' => ['checked' => TRUE],
+        ],
+      ],
     ];
 
     $defaults = $this->config(static::CONFIG_NAME)->get('2fa_conditions') ?? [];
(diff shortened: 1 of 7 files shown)


src/Form/AuthenticationSettingsForm.php
```diff
@@ -17,6 +24,25 @@ class AuthenticationSettingsForm extends ConfigFormBase {
    */
   const CONFIG_NAME = 'oe_authentication.settings';
 
+  /**
+   * The condition manager.
+   *
+   * @var \Drupal\Core\Executable\ExecutableManagerInterface&\Drupal\Core\Plugin\FilteredPluginManagerInterface
```
AaronGilMartinez, line 30: Nothing to fix here just an observation, it seems a bit odd the full namespaced classes with intersection types, extremely long.

AaronGilMartinez (changes requested):
Just one little remark

AaronGilMartinez (approved):

src/Form/AuthenticationSettingsForm.php
brummbar: Ye I know but there's a bit of a mess with the ConditionManager. There's no real interface that includes all the methods of the above.
But in truth, now that I think of it, we only need the method provided by FilteredPluginManagerInterface. I will drop it.

brummbar: Or maybe @donquixote has an opinion on this?

donquixote: So I see two separate questions here.
1. Code style and verbosity for union or instersection types in doc comments.
2. Whether we need this intersection type or can go with a single interface.
3. (not mentioned in your comments) Whether we can drop a `@var` doc type if it only repeats the native property type.

### For 1. - FQCN or alias

I see plenty examples of union types (`|`) in core which use full FQCN.

We could come up with a OpenEuropa rule to use aliases for union and intersection types.

The main problem with alias in doc comment is that either we would sometimes add an alias _only_ for the use in doc comments, or we have to make the doc comment conditional on whether the alias is already there for other reasons.

So for the sake of not losing sleep and time we should just put the FQCN.
(In my own side projects I have a mix of FQCN and alias, but this should not be relevant here)

### For 2. - all parts of intersection type needed?

In general we should document those interfaces that are needed where we use the object.
The doc type also has informational value, but this is mostly covered by the doc description.
In core I see a mix, where ExecutableManagerInterface is used in php param type, but at least in one place ConditionManager (class) is used in the property `@var` doc.

There is this issue in core, [2385427: Add a ConditionManagerInterface
](https://www.drupal.org/project/drupal/issues/2385427) which is open since a while.

---

### For 3. - Can we drop `@var` doc?

On the [doc standards page](https://www.drupal.org/docs/develop/standards/php/api-documentation-and-comment-standards#var) we find

> Typed class properties may omit the @var declaration. It is recommended to use typed properties whenever possible.

Seems it was introduced here: [3238192: Allow omitting @var for strictly typed class properties
](https://www.drupal.org/project/coding_standards/issues/3238192)

Also interesting: [3376518: Allow omitting @var, @param and @return tags or their descriptions if the variable name and type is enough
](https://www.drupal.org/project/coding_standards/issues/3376518)

AaronGilMartinez (approved):

PR #206, opened 2025-02-10
OPES-1549: Allow to delete users

### LINKED ISSUE #184: Delete user
The client has asked us to delete (not just block) a certain user. 
We are using oe_authentication and we have seen that it blocks this possibility. 
We wanted to confirm with you if this is a security restriction or if we can create a patch to remove that limitation?
Please can you explain the security reasons for this limitation?
Thanks!

AaronGilMartinez:
Fixes #184


oe_authentication.permissions.yml
```diff
@@ -1,3 +1,6 @@
 administer authentication configuration:
   title: 'Administer Authentication configuration'
   restrict access: false
+delete users:
+  title: 'Allows the user to use delete user options in "User cancel form"'
+  restrict access: true
```
donquixote, line 6: I feel uneasy about introducing a permission with such a generic name that could easily clash with permissions introduced by other modules or by future Drupal core versions.

If this has been agreed with others, then ok.

oe_authentication.module
```diff
@@ -13,7 +13,16 @@ declare(strict_types=1);
  * Remove the option to delete users while keeping the option to block them.
```
donquixote, line 13: This line needs to change now.
Also, to follow conventions it would say "Removes..." instead of "Remove...".

I like "permanently delete" to make it more clear. With this, I think we no longer need to mention "while keeping the option to block them". But can be discussed.

```suggestion
 * Removes the option to permanently delete users, depending on settings and permissions.
```
or
```suggestion
 * Conditionally removes the option to permanently delete users.
```

oe_authentication.module
```diff
@@ -13,7 +13,16 @@ declare(strict_types=1);
...
 function oe_authentication_user_cancel_methods_alter(&$methods) {
-  if (\Drupal::currentUser()->id() == 1) {
+  $current_user = \Drupal::currentUser();
+  $disabled_user_delete = \Drupal::configFactory()
+    ->get('oe_authentication.settings')
+    ->get('disable_user_delete');
+
+  if (
+    $disabled_user_delete !== TRUE ||
+    $current_user->id() == 1 ||
```
donquixote, line 23: Doesn't `->hasPermission()` already cover the user 1 case?
I mean it returns true for user 1 for all perms, right?

donquixote (changes requested):
First batch of feedback.
I want to submit this already, but I still want to have a closer look at the new tests.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -58,6 +58,11 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#title' => $this->t('Force two factor authentication'),
       '#default_value' => $this->config(static::CONFIG_NAME)->get('force_2fa'),
     ];
+    $form['disable_user_delete'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Disable the user delete options'),
+      '#default_value' => $this->config(static::CONFIG_NAME)->get('disable_user_delete'),
+    ];
```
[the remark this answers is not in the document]
donquixote: When I think about it more, wouldn't this setting belong into "Account settings" form, next to "When canceling a user account"?

tests/features/user-management.feature
```diff
@@ -1,13 +1,44 @@
...
   Scenario: Users should not be able to delete other users
     Given I am logged in as a user with the "administer users" permissions
     When I visit my user page
     And I click "Edit"
     And I click "Cancel account"
     Then I should see "Disable the account and keep its content."
     And I should not see "Delete the account and its content."
+
+  Scenario: As a user with permission "delete users" I can access the options
+    Given I am logged in as a user with the "administer users, delete users" permissions
```
donquixote, line 19: We are only testing delete own here, so we should not grant the "administer users" permission.
Instead, give "Cancel own user account", and "Select method for cancelling account".

tests/features/user-management.feature
```diff
@@ -1,13 +1,44 @@
...
+    When I visit my user page
+    And I click "Edit"
+    And I click "Cancel account"
+    Then I should see "Disable the account and keep its content."
+    And I should see "Delete the account and its content."
```
donquixote, line 24: I think we should cover all 4 methods, whether they are available or not.

tests/features/user-management.feature
donquixote: In fact the pre-existing scenario has the same problem.

tests/features/user-management.feature
donquixote: In a behat test can we just assert all options? E.g. the options for a field are A, B, C. So that, if option D appears, we get a fail.
I don't like the "I should not see ..." because this can lead to blind spots if we misspell the (not) expected text.

tests/features/user-management.feature
```diff
@@ -1,13 +1,44 @@
...
+  Or the module configuration to manage the delete options is disabled
 
-  @DrupalLogin
   Scenario: Users should not be able to delete other users
     Given I am logged in as a user with the "administer users" permissions
     When I visit my user page
     And I click "Edit"
     And I click "Cancel account"
     Then I should see "Disable the account and keep its content."
     And I should not see "Delete the account and its content."
```
donquixote, line 16: So, this case has the same problem as below, we only test with "administer users" permission but not with "cancel own.." and "select method for cancelling..".

Also we only cover the absence of "Delete the account and its content." but we miss to also cover the absence of "Delete the account and make its content belong to the %anonymous-name user."

(also see below about the fragility of testing the absence of something)

donquixote: Btw we can cover two "should not see" in one go if we only specify a partial phrase "Delete the account".

tests/Behat/AuthenticationContext.php
```diff
@@ -115,6 +115,22 @@ public function visitOwnUserPage(): void {
...
+  /**
+   * Navigates to other user's page.
+   *
+   * @Given I visit the :user_name user page
+   */
+  public function visitUserPage(string $user_name): void {
+    /** @var \Drupal\user\Entity\User $user */
+    $user = user_load_by_name($user_name);
+    if ($user === FALSE) {
+      throw new \InvalidArgumentException(sprintf('User with name %s could not be found.', $user_name));
```
donquixote, line 127: In the behat contexts in Drupal core I don't see "InvalidArgumentException" used, and only very few RuntimeException. Instead I see regular exception or ->assert*() calls.

To me, InvalidArgumentException and RuntimeException should indicate clear programming errors, but here the failure is conditional on the application state and data.

tests/features/configure_authentication.feature
```diff
@@ -25,6 +26,7 @@ Feature: Authentication
     And I fill in "Application assurance levels" with "assurance"
     And I fill in "Application available ticket types" with "ticket.test"
     And I check the box "Force two factor authentication"
+    And I uncheck the box "Disable the user delete options"
```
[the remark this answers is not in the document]
AaronGilMartinez: True

oe_authentication.permissions.yml
AaronGilMartinez: The name seems a bit generic indeed.
Since the permission is making available the options, maybe something related to what is doing would make more sense and hard to clash, like: 
- use delete user options
- delete user options access
- access delete user options

oe_authentication.module
AaronGilMartinez: Any of the two is ok for me.
About 'permanently' I'm not sure about it, is not that you can delete temporarily. 
The entity is deleted or not, seems more a problem of the options themselves that aren't clear enough.

oe_authentication.module
AaronGilMartinez: Since I want to keep functionality as is, I think is better to check the permission directly.

oe_authentication.post_update.php
```diff
@@ -42,3 +42,12 @@ function oe_authentication_post_update_00004(): void {
...
 }
+
+/**
+ * Set the disable_user_delete to TRUE as default.
+ */
+function oe_authentication_post_update_00005(): void {
+  \Drupal::configFactory()->getEditable('oe_authentication.settings')
+    ->set('disable_user_delete', TRUE)
+    ->save();
+}
```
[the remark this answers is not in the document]
AaronGilMartinez: The module relied on update_hooks for other configuration, I did not want to add it just to this configuration, feels something should be done for all the configuration keys but is not in the scope.

Yes, I guess that setting could be inverted, would mean no path update. I'm considering it-

src/Form/AuthenticationSettingsForm.php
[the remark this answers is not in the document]
AaronGilMartinez: The module uses the checkbox for other boolean fields, like 2fa authentication.
And for the place, since it's something the module it's doing seems the right place to add it.

oe_authentication.module
```diff
@@ -13,7 +13,16 @@ declare(strict_types=1);
...
+    $current_user->hasPermission('delete users')
+  ) {
     return;
   }
```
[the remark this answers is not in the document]
AaronGilMartinez: I don't like the idea of modifing/refactoring the current code/behaviour just the minimum to allow the deletion.

tests/features/user-management.feature
AaronGilMartinez: I understand, ideally we would assert that there are only 3 options and that they are the right ones. 
No need of checking not available options.
Not sure if behat provides  steps for this, I check, if not it will mean custom steps that I prefer to avoid.

tests/features/user-management.feature
AaronGilMartinez: I would say yes, they are two paths to get to the same destination.
Inded

tests/features/user-management.feature
AaronGilMartinez: Indeed

oe_authentication.module
AaronGilMartinez: I discard the direct comparisson, inverting the option seems a better idea.
We have two options to allow:
 - When option is enabled or user has the permission. 
   - Problem here is that we are granting access to other users than 1: admin users.
 - When user is superuser or (option is enabled and has the permission).
   - This requires two conditions for the users to get access to this options.

oe_authentication.permissions.yml
AaronGilMartinez: Renamed to: access user delete options
Describes what it does and has a similar name to the configuration which helps to relate between each other.
It's more concise to avoid collision with other possible names.

config/schema/oe_authentication.schema.yml
```diff
@@ -20,3 +20,6 @@ oe_authentication.settings:
     force_2fa:
       type: boolean
       label: 'Force two factor authentication'
+    disable_user_delete:
+      type: boolean
+      label: 'Disable the user delete options'
```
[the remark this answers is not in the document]
AaronGilMartinez: Inverted the option so the name has changed totally.

oe_authentication.permissions.yml
```diff
@@ -1,3 +1,6 @@
 administer authentication configuration:
   title: 'Administer Authentication configuration'
   restrict access: false
+delete users:
+  title: 'Allows the user to use delete user options in "User cancel form"'
```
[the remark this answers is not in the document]
AaronGilMartinez: Used first one.

oe_authentication.post_update.php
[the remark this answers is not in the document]
AaronGilMartinez: Inverted the config at the end so no update path is needed, and in case someone skips the hook won't break anything.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -58,6 +58,11 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       '#title' => $this->t('Force two factor authentication'),
       '#default_value' => $this->config(static::CONFIG_NAME)->get('force_2fa'),
     ];
+    $form['disable_user_delete'] = [
+      '#type' => 'checkbox',
+      '#title' => $this->t('Disable the user delete options'),
```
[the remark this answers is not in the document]
AaronGilMartinez: Title no longer applies, but I like the description.

tests/features/user-management.feature
AaronGilMartinez: Added checks for all the options, I think the best for the negative assertions is to use "Delete the account" as you suggest.

tests/Behat/AuthenticationContext.php
donquixote: Ok, tbh I am not sure what is the correct thing to throw here, just going by existing examples.

oe_authentication.permissions.yml
donquixote: Ok, I leave this open so that @brummbar could add his thoughts.
The new name ties this to the UI and the options in the user cancel form, which would imply that other ways to permanently delete a user would need a different permission.
But let's keep it as is until we get a 3rd opinion.

oe_authentication.module
donquixote: The "permanently" is just to distinguish from "cancel user", to make it clear that "cancel user" is a distinct option and so that the distinction is clear, which is why I considered the word "permanently"...

But yeah I think just "delete" is fine.

tests/Behat/AuthenticationContext.php
AaronGilMartinez: Module already provides steps with some exceptions, added Exception to keep consistency.

oe_authentication.module
```diff
@@ -10,10 +10,19 @@ declare(strict_types=1);
...
+  $current_user = \Drupal::currentUser();
+  $restrict_user_delete = \Drupal::configFactory()
+    ->get('oe_authentication.settings')
+    ->get('restrict_user_delete_cancel_methods');
+  // @todo The restriction should not check for UID 1 as it's bad practice. It
+  // should either check for the user being an admin, or use a permission based
+  // approach like the module drupal/user_cancel_methods_permissions.
+  // Furthermore, the 'access' property of the cancel methods should be used,
+  // rather than removing the method altogether.
+  if ($current_user->id() == 1 || $restrict_user_delete !== TRUE) {
```
donquixote, line 25: Not sure if we need `!== TRUE`, if the config value is boolean.
It's a matter of trust, convention and expressiveness.
Trust that we always get boolean thanks to the config schema.
Expressiveness in showing that yes we expect a boolean TRUE here.

I say keep it (because I don't have strong opinion otherwise).

oe_authentication.post_update.php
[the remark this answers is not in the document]
donquixote: well that last comment no longer applies.

(My above suggestion is an example for putting too many ideas on the table.)

src/Form/AuthenticationSettingsForm.php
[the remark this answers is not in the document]
donquixote: Changed according to new agreement.

tests/Behat/AuthenticationContext.php
```diff
@@ -115,6 +115,28 @@ public function visitOwnUserPage(): void {
...
+   * @var string $username
+   *   The name of the user whose page is to be visited.
+   *
+   * @Given I visit the :user_name user page
+   *
+   * @throws \Exception
+   *   Thrown when the user with the given name does not exist.
+   */
+  public function visitUserPage(string $user_name): void {
+    /** @var \Drupal\user\Entity\User $user */
```
donquixote, line 130: This would include FALSE at this step in the method.
Only after the `if (..)` the FALSE goes away.
```suggestion
    /** @var \Drupal\user\Entity\User|false $user */
```
However, it seems we don't really need this at all.
The `user_load_by_name()` is documented to return `UserInterface|false`, which is good enough for everything we do with that variable.

donquixote: so
```suggestion
```

tests/features/user-management.feature
```diff
@@ -1,13 +1,41 @@
...
-  @DrupalLogin
   Scenario: Users should not be able to delete other users
     Given I am logged in as a user with the "administer users" permissions
     When I visit my user page
     And I click "Edit"
     And I click "Cancel account"
     Then I should see "Disable the account and keep its content."
-    And I should not see "Delete the account and its content."
+    And I should see "Disable the account and unpublish its content."
+    And I should not see "Delete the account"
```
donquixote, line 16: ok, this one covers both options, good.

tests/Behat/AuthenticationContext.php
```diff
@@ -143,4 +165,28 @@ public function allowExternalUsers(): void {
     $this->configContext->setConfig('oe_authentication.settings', 'assurance_level', 'LOW');
   }
 
+  /**
+   * Configures oe_authentication module to allow to use user delete options.
+   *
+   * @Given the site is configured to display user delete options
```
donquixote, line 171: We _could_ call this "is configured to not restrict user delete options".
But to me it is not a blocker.

donquixote: so it would be
```suggestion
   * @Given the site is configured to not restrict user delete options
```

oe_authentication.module
```diff
@@ -10,10 +10,19 @@ declare(strict_types=1);
...
-  if (\Drupal::currentUser()->id() == 1) {
+  $current_user = \Drupal::currentUser();
+  $restrict_user_delete = \Drupal::configFactory()
+    ->get('oe_authentication.settings')
+    ->get('restrict_user_delete_cancel_methods');
+  // @todo The restriction should not check for UID 1 as it's bad practice. It
+  // should either check for the user being an admin, or use a permission based
+  // approach like the module drupal/user_cancel_methods_permissions.
+  // Furthermore, the 'access' property of the cancel methods should be used,
+  // rather than removing the method altogether.
```
donquixote, line 24: This second `@todo` could be moved below into the foreach().

donquixote (changes requested):
some changes suggested, but not hard blockers.

donquixote:
Some points we missed in our latest planning.

### Conflict with default method
If we enable "Restrict options" setting, but set the default cancel method to something that deletes, then:
- Users with option to choose a cancel method can only choose the remaining methods.
- Users without that option, but permission to cancel their own account, can permanently delete their own account.

### Which form to use
From UX perspective it seems this setting belongs into "Account settings" form, not "Authentication settings".
Only from the module developer perspective it would make sense to put it in "Authentication settings" because the functionality is provided by oe_authentication module.

TBD with rest of the team.

brummbar:
As per https://github.com/openeuropa/oe_authentication/issues/79#issuecomment-501325503, it's for the site developers to configure a proper default method when restricting the options. Since the module behaves already in that way, we won't add anything as part of this pull request. We could iterate in a different one.

oe_authentication.module
brummbar, line 24: Lines after the first need to be indented
```suggestion
  // @todo The restriction should not check for UID 1 as it's bad practice. It
  //   should either check for the user being an admin, or use a permission based
  //   approach like the module drupal/user_cancel_methods_permissions.
  //   Furthermore, the 'access' property of the cancel methods should be used,
  //   rather than removing the method altogether.
```

oe_authentication.module
```diff
@@ -10,10 +10,19 @@ declare(strict_types=1);
...
  *
- * Remove the option to delete users while keeping the option to block them.
+ * Conditionally removes the option to delete users.
  */
 function oe_authentication_user_cancel_methods_alter(&$methods) {
-  if (\Drupal::currentUser()->id() == 1) {
+  $current_user = \Drupal::currentUser();
+  $restrict_user_delete = \Drupal::configFactory()
+    ->get('oe_authentication.settings')
+    ->get('restrict_user_delete_cancel_methods');
```
brummbar, line 19: We forgot to add a fallback to TRUE, just for the sake of it.

tests/features/user-management.feature
```diff
@@ -1,13 +1,41 @@
...
+    When I visit the foo user page
+    And I click "Edit"
+    And I click "Cancel account"
+    Then I should see "Disable the account and keep its content."
+    And I should see "Disable the account and unpublish its content."
+    And I should see "Delete the account and its content."
+    And I should see "Delete the account and make its content belong to the Anonymous user."
+
+  Scenario: Users should be able to delete other users when the configuration to
+  restrict the user delete options is not enabled.
```
brummbar, line 32: Don't split this

brummbar (changes requested):

oe_authentication.module
donquixote: I thought we don't, and rely on upgrade path?
I don't mind if we add it..

oe_authentication.module
donquixote: I see it has been indented now.

donquixote: This has been done.

oe_authentication.module
brummbar: I don't know we went back and forth a lot on this. I first said no, then I said ok to add it just to avoid any disruption.

brummbar (approved):

donquixote (approved):

sergepavle (approved):

nagyad (approved):

PR #211, opened 2025-02-21
OPES-1548: Allow developers to customise the message shown when a login is denied for missing 2FA

### LINKED ISSUE #205: Allow developers to customise the message shown when a login is denied for missing 2FA
Allow to customise in the message shown when a login is rejected due to missing 2FA.

Make sure the message is translatable.

brummbar:
Fixes #205 .

### DIFF
--- a/config/schema/oe_authentication.schema.yml
+++ b/config/schema/oe_authentication.schema.yml
@@ -28,3 +28,6 @@ oe_authentication.settings:
       sequence:
         type: condition.plugin.[id]
         label: 'Two-factor authentication condition'
+    message_login_2fa_required:
+      type: text
+      label: 'Error message for login denied: two-factor authentication required'
--- a/src/Event/TwoFactorAuthenticationEventSubscriber.php
+++ b/src/Event/TwoFactorAuthenticationEventSubscriber.php
@@ -82,7 +82,14 @@ class TwoFactorAuthenticationEventSubscriber implements EventSubscriberInterface
     catch (\Throwable $exception) {
       // If any exception happens, we cannot trust the login attempt anymore.
       // Use the default error message from the CAS module.
-      Error::logException($this->logger, $exception);
+      Error::logException(
+        $this->logger,
+        $exception,
+        'An exception occurred when evaluating 2FA conditions for account with uid @uid. ' . Error::DEFAULT_ERROR_MESSAGE,
+        [
+          '@uid' => $event->getAccount()->id(),
+        ],
+      );
       $event->cancelLogin($this->casHelper->getMessage('error_handling.message_validation_failure'));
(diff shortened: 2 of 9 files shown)


tests/Functional/TwoFactorAuthenticationTest.php
```diff
@@ -207,6 +207,15 @@ public function testTwoFactorAuthenticationConditions(): void {
     $this->drupalLogout();
     $this->casLogin('basic_user@example.com', 'pwd1');
     $assert_session->statusMessageContains('You are required to log in using a two-factor authentication method.', 'error');
```
AaronGilMartinez, line 209: We could store this text into a variable to assert later that is not present. 
The  assertion to check that the message is not present will pass anyway if there is a missalignment between these texts.

src/Event/TwoFactorAuthenticationEventSubscriber.php
```diff
@@ -75,52 +76,59 @@ public function checkTwoFactorAuthentication(CasPreLoginEvent $event) {
...
+    $conditions_configuration = $config->get('2fa_conditions') ?? [];
+    try {
+      if ($this->isTwoFactorAuthenticationRequiredForUser($event->getAccount(), $conditions_configuration)) {
+        $event->cancelLogin($config->get('message_login_2fa_required'));
+      }
+    }
+    catch (\Throwable $exception) {
+      // If any exception happens, we cannot trust the login attempt anymore.
+      // Use the default error message from the CAS module.
+      Error::logException($this->logger, $exception);
```
AaronGilMartinez, line 88: Since we are logging the exception I would add a little check to the test to verify it.

AaronGilMartinez (changes requested):

config/schema/oe_authentication.schema.yml
```diff
@@ -28,3 +28,6 @@ oe_authentication.settings:
       sequence:
         type: condition.plugin.[id]
         label: 'Two-factor authentication condition'
+    message_login_2fa_required:
```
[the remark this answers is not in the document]
brummbar: Text is already translatable https://www.drupal.org/docs/drupal-apis/configuration-api/configuration-schemametadata#types
I did not test for it, but manually, as it's overkill to write tests for that.

src/Event/TwoFactorAuthenticationEventSubscriber.php
brummbar: I am not sure how critical is to do that, I test for the behaviour mostly. I will check if core does so.
We did it often in our modules, but it's also cumbersome to do.

AaronGilMartinez (approved):

PR #213, opened 2025-02-27
OPES-1561: Improve UI/UX for 2FA settings in configuration form

### LINKED ISSUE #212: Improve UI/UX for 2FA settings in configuration form
The current UX when configuring the 2FA on oe_authentication is confusing.

It is not clear if a condition is enabled, and what's the status of it.
This can lead to scenarios where it seems that the 2FA has been disabled for all users, but actually there is a condition configured that will enforce it for some.
The only way to determine if a condition is enabled is to check the configuration exported.

The scope of this ticket is:

1. Add a radios element with the label "Use two-factor authentication" and options "Always", "Never", "Conditional".
2. Maintain the value of the current force_2fa setting for backwards compatibility. Map "Always" to TRUE, the rest to FALSE.
3. When "Conditional" is selected, the conditions show. For each condition, add a checkbox that clearly shows if they are enabled or not.
4. When a condition is enabled, the related configuration form is shown.

brummbar:
Fixes #212


tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -0,0 +1,273 @@
...
+  /**
+   * Tests the two-factor authentication condition settings.
+   */
+  public function testTwoFactorAuthenticationConditionSettings(): void {
+    $role_id = $this->createRole(['administer authentication configuration']);
+    $this->drupalLogin($this->createUser(values: ['roles' => [$role_id]]));
+    $this->drupalGet('/admin/config/system/oe_authentication');
+
+    $assert_session = $this->assertSession();
+    $use_2fa_fieldset = $assert_session->elementExists('named', ['fieldset', 'Use two-factor authentication']);
```
donquixote, line 39:
```suggestion
    $use_2fa_radios_fieldset = $assert_session->elementExists('named', ['fieldset', 'Use two-factor authentication']);
```
If you don't see the form, you can easily get confused if this is just for the radios or some kind of outer wrapper for everything related to 2fa.

tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -0,0 +1,273 @@
...
+    $this->assertCount(3, $use_2fa_fieldset->findAll('css', 'input'));
+    $never_radio = $assert_session->fieldExists('Never', $use_2fa_fieldset);
+    $this->assertTrue($never_radio->isChecked());
+    $always_radio = $assert_session->fieldExists('Always', $use_2fa_fieldset);
+    $this->assertFalse($always_radio->isChecked());
+    $conditional_radio = $assert_session->fieldExists('Conditional', $use_2fa_fieldset);
+    $this->assertFalse($conditional_radio->isChecked());
```
donquixote, line 46: I wish we could assert radios more easily. But ok.

tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -0,0 +1,273 @@
...
+    // The "always" option sets the "force_2fa" to TRUE.
+    $config = $this->loadConfig();
+    $this->assertEquals([], $config->get('2fa_conditions'));
+    $this->assertTrue($config->get('force_2fa'));
+
+    $this->drupalGet('/admin/config/system/oe_authentication');
+    $this->assertFalse($conditions_wrapper->isVisible());
+    $conditional_radio->click();
+    $conditions_wrapper_selector = 'details[data-drupal-selector="edit-2fa-conditions"]';
+    $conditions_wrapper = $assert_session->waitForElementVisible('css', $conditions_wrapper_selector);
```
donquixote, line 76: What's a bit interesting: Here we wait until it is visible.
But above when we assert it is not visible after a click, we don't wait.
So in theory it could become visible above 10 milliseconds after clicking "Always".

(There is nothing we can do here, just an observation. You can resolve.)

tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -0,0 +1,273 @@
...
+    $this->assertFalse($assert_session->fieldExists('Example configuration option', $test_condition_wrapper)->isChecked());
+    // The second one is the user role core plugin.
+    $user_role_condition_wrapper = $this->assertDetailsElementExists('User Role', $condition_settings_tabs_wrapper);
+    $this->assertEquals([
+      'anonymous',
+      'authenticated',
+      $role_id,
+    ], array_map(
+      function (NodeElement $element): string {
+        $this->assertFalse($element->isChecked());
```
donquixote, line 104: I don't like to see this assert in the closure for array_map(). But it might be the most efficient way to do it. Ok.

donquixote (commented):
I am not finished yet with the review of the test.
But I want to submit already so that you are not blocked.

Overall it looks good, and most of my feedback is optional stuff that we could do later or never.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -174,30 +204,46 @@ protected function getUserConditionDefinitions(): array {
...
-        SubformState::createForSubform($form['2fa_conditions'][$condition_id], $form, $form_state),
+        $form['2fa_conditions']['settings'][$condition_id],
+        SubformState::createForSubform($form['2fa_conditions']['settings'][$condition_id], $form, $form_state),
       );
     }
+
+    // When the conditions mode is active, at least one condition must be set.
+    if (empty($enabled_conditions) && $form_state->getValue('use_2fa') === 'conditions') {
+      $form_state->setError($form['2fa_conditions']['status'], $this->t('At least one condition should be enabled when two-factor authentication is set to conditional.'));
+    }
```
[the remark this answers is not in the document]
brummbar: We already discussed this during a call. The reason for not doing it is that I don't want to call the submitConfiguration of the plugins in the validate.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -52,46 +53,69 @@ public function getFormId() {
...
+      '#type' => 'details',
+      '#open' => TRUE,
+      '#title' => $this->t('Two-factor authentication conditions'),
+      '#states' => [
+        'visible' => [
+          ':input[name="use_2fa"]' => ['value' => 'conditions'],
+        ],
+      ],
+    ];
+    $form['2fa_conditions'] = $this->buildTwoFactorConditionsInterface($form['2fa_conditions'], $form_state);
```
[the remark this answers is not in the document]
brummbar: + operator won't allow to replace items already existing. It's not a blocker to have it like this, and it's a common pattern, so I'll leave it as it is.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -216,20 +262,39 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
       $collection->addInstanceId($condition_id, $condition_configuration);
+
+      // Keep the label of the plugin since we have it instantiated already.
+      $plugin_labels[$condition_id] = $condition->getPluginDefinition()['label'];
+    }
+
+    $configuration = $collection->getConfiguration();
+    $default_config_plugins = array_diff_key($plugin_labels, $configuration);
+    if (empty($configuration) && $default_config_plugins) {
+      $this->messenger->addWarning('All condition plugins have been disabled, as no configuration was provided. Two-factor authentication has been set to "never".');
```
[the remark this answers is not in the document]
brummbar: I'll capitalise the N, but the message is correct. We already made a point in the text that no configuration = default configuration.

tests/FunctionalJavascript/SettingsFormTest.php
brummbar: It's a test. Not important.

tests/FunctionalJavascript/SettingsFormTest.php
brummbar: To be honest the wait it's not really needed as there's no animation, and JS tests have implicit waits.
But the main reason for not waiting for the method to be hidden is that I don't have an helper ready.

src/Form/AuthenticationSettingsForm.php
```diff
@@ -216,20 +262,39 @@ public function submitForm(array &$form, FormStateInterface $form_state) {
...
       );
 
       $condition_configuration = $condition->getConfiguration();
       $collection->addInstanceId($condition_id, $condition_configuration);
+
+      // Keep the label of the plugin since we have it instantiated already.
+      $plugin_labels[$condition_id] = $condition->getPluginDefinition()['label'];
+    }
+
+    $configuration = $collection->getConfiguration();
```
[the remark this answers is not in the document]
brummbar: This is explained already in the messages.

tests/FunctionalJavascript/SettingsFormTest.php
donquixote: the main problem is we would always have to wait the maximum time to verify that something does _not_ happen. So yeah nothing we can do, it just means the asserts above may not be full reliable.

AaronGilMartinez (changes requested):

tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -0,0 +1,273 @@
...
+    // required to be selected.
+    $assert_session->buttonExists('Save configuration')->press();
+    $assert_session->statusMessageContains('At least one condition should be enabled when two-factor authentication is set to conditional.', 'error');
+
+    // Test that configuration is required when enabling a condition.
+    $enabled_conditions_fieldset->checkField('User test condition');
+    $assert_session->buttonExists('Save configuration')->press();
+    $assert_session->statusMessageContains('All condition plugins have been disabled, as no configuration was provided. Two-factor authentication has been set to "never".', 'warning');
+    $assert_session->statusMessageContains('The configuration options have been saved.', 'status');
+    $assert_session->statusMessageNotExists('error');
```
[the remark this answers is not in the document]
brummbar: Since we send some messages during submit rather than validate, I wanted to make sure that no messages are shown additionally.

AaronGilMartinez (approved):

PR #215, opened 2025-03-03
OPES-1566: Prevent redirect loop when 2FA is not used but required by conditions

### LINKED ISSUE #214: Prevent redirect loop when 2FA is not used but required by conditions
When conditions are configured, if a user logs in without using a 2FA method, they will be requested to log in again using one.
If the user is already logged in EU Login, they will redirected back instantly without giving a chance to change the login method. 

Add the possibility to show a specially crafted URL that will send the user to EU Login allowing them to select the correct login method.
Add the possibility to use this URL in the message "Message for login rejected: two-factor authentication required".

brummbar:
Fixes #214


tests/Kernel/EuLoginEventSubscriberTest.php
```diff
@@ -91,12 +91,39 @@ public function test2faRedirectParameter(): void {
...
     $config->set('force_2fa', FALSE)->save();
     $response = $this->container->get('http_kernel')->handle(clone $request);
     $this->assertEquals(302, $response->getStatusCode());
     $this->assertStringNotContainsString('acceptStrengths', $response->getContent());
+
+    // The 2FA parameter is added if the page request has a "force_2fa" query
+    // parameter.
+    $request = Request::create(Url::fromRoute('user.login', options: [
+      'query' => [
+        'force_2fa' => TRUE,
```
AaronGilMartinez, line 105:
```suggestion
        'force_2fa' => 1,
```

AaronGilMartinez (changes requested):

AaronGilMartinez (approved):

bramtenhove (approved):
We tested the changes and works great. Thanks for the efforts!

PR #222, opened 2025-03-17
OPES-1576: Show only conditions that require a user

brummbar:


tests/FunctionalJavascript/SettingsFormTest.php
```diff
@@ -263,6 +264,16 @@ function (NodeElement $element): string {
     $config = $this->loadConfig();
     $this->assertFalse($config->get('force_2fa'));
     $this->assertEquals([], $config->get('2fa_conditions'));
+
+    // Since we rely on a test module from core, and we also test that the
+    // following plugins don't show in the UI, we won't detect if the plugins
+    // have been removed. Add a safety check to look for their existence.
+    // This doesn't prevent core from changing completely their context
+    // definitions, but that's more unlikely.
```
donquixote, line 272: This comment is ok, but as a reader I would wish to get reminded of the context before the finer details.
```suggestion
    // The following conditions are provided by the 'condition_test' module, but are not shown in the form, because they don't match the criteria.
    // The assertions allow us to detect if the plugins get removed in a future version of Drupal core.
    // They will not tell is if their context definition changes, but that is less likely.
```

src/Form/AuthenticationSettingsForm.php
```diff
@@ -184,17 +184,28 @@ protected function getUserConditionDefinitions(): array {
...
+
+      if (empty($context_definitions) || count($context_definitions) > 1) {
+        return FALSE;
       }
 
-      return FALSE;
+      // At this point the single context definition available is a user entity
+      // one. If it's required, the plugin is allowed.
+      return current($context_definitions)->isRequired();
     });
```
donquixote, line 209: So to be clear:\
We are filtering conditions in the form, but we don't filter them when they are applied, or the config is stored.\
This means they could be enabled by hacking the config yml and then importing.

I don't think it is a problem, site builders who edit config files do at their own risk.\
Also it would be out of scope to change it here.\
I only comment because I think it is the first time I look at this code (or maybe I don't remember it).

src/Form/AuthenticationSettingsForm.php
```diff
@@ -184,17 +184,28 @@ protected function getUserConditionDefinitions(): array {
...
-      if (is_array($definition) && isset($definition['context_definitions'])) {
-        return !empty($definition['context_definitions']);
+      elseif (is_array($definition) && isset($definition['context_definitions'])) {
+        $context_definitions = $definition['context_definitions'];
+      }
+      else {
+        return FALSE;
+      }
+
+      if (empty($context_definitions) || count($context_definitions) > 1) {
```
donquixote, line 202: Could be just `count(*) !== 1`? Or can `$context_definitions` be NULL?

src/Form/AuthenticationSettingsForm.php
```diff
@@ -184,17 +184,28 @@ protected function getUserConditionDefinitions(): array {
...
+    return array_filter($definitions, function ($definition): bool {
       // @see \Drupal\Core\Plugin\Context\ContextHandler::getContextDefinitions()
       if ($definition instanceof ContextAwarePluginDefinitionInterface) {
-        return !empty($definition->getContextDefinitions());
+        $context_definitions = $definition->getContextDefinitions();
       }
-      if (is_array($definition) && isset($definition['context_definitions'])) {
-        return !empty($definition['context_definitions']);
+      elseif (is_array($definition) && isset($definition['context_definitions'])) {
+        $context_definitions = $definition['context_definitions'];
```
donquixote, line 196: We _could_ do this:
```suggestion
      elseif (is_array($definition)) {
        $context_definitions = $definition['context_definitions'] ?? [];
```
We already check for `empty()` below.
But, there is some symmetry to skipping those without a context definition, it is symmetric to the instanceof above.

donquixote (approved):
All the comments are just thoughts, no need to change anything.

src/Form/AuthenticationSettingsForm.php
brummbar: Yes config can be manually changed, but that's the same everywhere.

Issue #229, opened 2025-05-21
Missing "langcode" in oe_authentication.settings


vever001:
## Problem
Release 1.20.0 introduced https://github.com/openeuropa/oe_authentication/pull/216 and added `message_login_2fa_required` in `oe_authentication.settings` which is a translatable config of type text. Now config_inspector (via `drush config:inspect `command) throws the following:
```
oe_authentication.settings   Correct   57%           1 errors   ValidKeys: '<infer>'                        
                                                                  LangcodeRequiredIfTranslatableValues: null  
                                                                  FullyValidatable: null
```

Related change record (introduced in 10.3.0): https://www.drupal.org/node/3427629

## Steps to reproduce
Execute: `drush config:inspect --only-error --detail --list-constraints`

## Proposed resolution
Add missing langcode: en in oe_authentication.settings + add post update hook.

allternativ9:
Dear @vever001 ,
I have created a ticket in our board to review the issue and the proposed PR. We'll keep you posted on the progress.

Many thanks,
Angela.

PR #230, opened 2025-05-21
Issue #229: Missing "langcode" in oe_authentication.settings.

vever001:
See https://github.com/openeuropa/oe_authentication/issues/229

### DIFF
--- a/config/install/oe_authentication.settings.yml
+++ b/config/install/oe_authentication.settings.yml
@@ -1,3 +1,4 @@
+langcode: en
 protocol: 'eulogin'
 register_path: 'eim/external/register.cgi'
 validation_path: 'TicketValidationService'
--- a/oe_authentication.post_update.php
+++ b/oe_authentication.post_update.php
@@ -61,3 +61,12 @@ function oe_authentication_post_update_00006(): void {
     ->set('message_login_2fa_required', 'Your account is required to log in using a two-factor authentication method. Please <a href=":login">log in again via this link</a>.')
     ->save();
 }
+
+/**
+ * Add missing langcode.
+ */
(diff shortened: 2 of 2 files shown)


config/install/oe_authentication.settings.yml
```diff
@@ -1,3 +1,4 @@
+langcode: en
```
donquixote, line 1: This fix makes sense.
I do find similar changes in core, and it aligns perfectly with the change record.

oe_authentication.post_update.php
```diff
@@ -61,3 +61,12 @@ function oe_authentication_post_update_00006(): void {
...
 }
+
+/**
+ * Add missing langcode.
+ */
+function oe_authentication_post_update_00007(): void {
+  \Drupal::configFactory()->getEditable('oe_authentication.settings')
+    ->set('langcode', 'en')
+    ->save();
+}
```
donquixote, line 72: This change makes sense to me.\
But, I have not found similar update hooks in core (I searched for `->set('langcode', 'en')`.\
So, what do we miss if we don't do this?

Also, is this a case for hook_update_N() or rather hook_post_update_NAME()?

Going to think about this or ask others..

donquixote: One thing I was wondering: Do we need to worry about sites that already saved this with e.g. `langcode: fr`?

I don't think this is a problem, because:
- The addition of the text value is very recent.
- The locale module does add the site default language to a config record that has translatable strings and no langcode. But this is actually something we want to fix, because we want 'en' here.

donquixote (approved):

donquixote (approved):

AaronGilMartinez (approved):

